PR #9, opened 2022-03-11
Change activated level comparison method

mircobabini:
`pmpro_hasMembershipLevel(x)`  does not permit to check for membership levels that are not activated after checkout (like gifts). Must use `$pmpro_level->id` instead.

Should also update: https://www.paidmembershipspro.com/set-up-unique-membership-confirmation-pages-based-on-level/


dparker1005 (approved):
This change also helps to better support MMPU.

PR #11, opened 2022-04-09
Add PayPal IPN Forwarder

mircobabini:
Started from https://gist.github.com/andrewlimaza/3864c58c0cf1c32ece2e381376ba6ca5, but changed in some ways to be more secure, timed and looking very similar to the real ipn requests from PayPal.

Raw idea from https://www.angelleye.com/product/wordpress-paypal-ipn-forwarder/


andrewlimaza:
I can't pull this into GH, closing the issue and reworking it slightly.

PR #15, opened 2022-05-25
Add membership level name to expired email template

ipokkel:
Retrieve the last expired membership level for a user and add `!!membership_level_name!!` as an email variable.

Original gist: https://gist.github.com/ipokkel/dde134d098be8b5b6d3a7e63c79d84cc


andrewlimaza:
This has been merged in from another PR :) (Identical)

PR #23, opened 2022-06-08
Code to Add Google Tag Manager to Your Website Body

SamDaugherty4:
Code to Add Google Tag Manager to Your Website Body adding to snippet library


andrewlimaza:
I'm thinking that these code snippets can stay on gist.github.com (especially that they are separate snippets that are needed to go into one solution). The snippets library goal is to have single copy/paste snippets and this may be confusing to some that stumble upon the snippets without context.

PR #37, opened 2022-08-09
Adding gists for upcoming tooltips

kimcoleman:
Adding a few gists to be used in upcoming tooltips.

- Show next and last payment date on members list and members list CSV export.
- Show renewal date on orders list
- Remove one or unset all level templates from the Add New membership level pop up in PMPro v2.9+
- Add new level template 'quarterly' to popup and create defaults for level object
- Function to retrieve active members of a level with required usermeta field.


andrewlimaza (approved):

andrewlimaza:
Thanks for this! Some cool new gists.

PR #60, opened 2022-08-23
Adding gist for Add a conditional text field for referral name if a b…

kimcoleman:
…uilt-in select field named 'how_hear' is set to 'other'.

### DIFF
--- a/user-fields/add-conditional-referred-by-user-field.php
+++ b/user-fields/add-conditional-referred-by-user-field.php
@@ -21,6 +21,25 @@ function my_pmpro_add_referral_source_conditional_user_field() {
 	// Store our field settings in an array.
 	$fields = array();
 	
+	// The how_hear dropdown. If it wasn't added via settings, use this code to add.
+	/*
+	$fields[] = new PMPro_Field(
+		'how_hear',
+		'select',
+		array(
+			'label'				=> 'How did you hear about us?',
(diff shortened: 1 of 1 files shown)


user-fields/add-conditional-referred-by-user-field.php
```diff
@@ -0,0 +1,44 @@
...
+function my_pmpro_add_referral_source_conditional_user_field() {
+	// Don't break if PMPro is out of date or not loaded.
+	if ( ! function_exists( 'pmpro_add_user_field' ) ) {
+		return false;
+	}
+
+	// Store our field settings in an array.
+	$fields = array();
+	
+	// The conditional field that only shows if the field name 'how_hear' is set to 'other'.
```
ideadude, line 43:
```suggestion
	// The how_hear dropdown. If it wasn't added via settings, use this code to add.
	/*
	$fields[] = new PMPro_Field(
		'how_hear',
		'select',
		array(
			'label'				=> 'How did you hear about us?',
			'profile'			=> 'admins',
			'memberslistcsv'	=> true,
			'options'			=> array(
									'Twitter',
									'Facebook',
									'A Friend',
									'Other',
								)
		)
	);
	*/
	
	// The conditional field that only shows if the field name 'how_hear' is set to 'other'.
```

PR #71, opened 2022-11-19
Adding recipe to set category products to free for members only.

kimcoleman:
Adding snippet to demo the pmprowoo_get_membership_price filter and setting a certain product category to free for anyone with a membership level.

### DIFF
--- a/add-ons/pmpro-woocommerce/set-ebook-category-member-price-to-free.php
+++ b/add-ons/pmpro-woocommerce/set-ebook-category-member-price-to-free.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * This recipe sets all products in the Ebooks category to free for logged in members.
+ *
+ * title: Set products in Ebook category to free for members only.
+ * layout: snippet
+ * collection: add-ons
+ * category: pmpro-woocommerce
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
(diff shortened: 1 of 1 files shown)


add-ons/pmpro-woocommerce/set-ebook-category-member-price-to-free.php
```diff
@@ -0,0 +1,31 @@
...
+		return $discount_price;
+	}
+
+	// Check if the product is in the Ebook category and set price to 0 for all members.
+	if ( has_term( array( 'ebooks' ), 'product_cat', $product->get_id() ) ) {
+		$discount_price = 0;
+	}
+
+	// Return the discounted price.
+    return $discount_price;
```
dparker1005:
```suggestion
	return $discount_price;
```

dparker1005 (approved):
Quick spacing fix, but looks good!

add-ons/pmpro-woocommerce/set-ebook-category-member-price-to-free.php
```diff
@@ -0,0 +1,31 @@
...
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+function my_pmprowoo_get_membership_price_ebooks_category( $discount_price, $level_id, $original_price, $product ) {
+	// Return early if the user does not have a membership level.
+	// Adjust this or add additional checks to require a specific membership level.
+	if ( function_exists( 'pmpro_hasMembershipLevel' ) && ! pmpro_hasMembershipLevel() ) {
+		return $discount_price;
+	}
```
dparker1005, line 21: Actually, thinking about this again, users will always have a membership level or be buying one when this filter is used. That level is passed in through $level_id. Maybe a better example would be "level 1 users get free e-books". Adding the change here in case we want to go this route.
```suggestion
	// We only want to make e-books free if the user has (or is buying) level 1.
	if ( $level_id != 1 ) {
		return $discount_price;
	}
```

PR #78, opened 2023-01-16
Add Percent off Discount Codes

kimwhite:
Option to create percent off discount codes. From Hersha https://gist.github.com/greathmaster/54e2e0e60f875af3efbd25b899812e6e


JarrydLong (approved):

dparker1005:
Right now, the discount affects both initial and recurring prices, but many sites (including paidmembershipspro.com) only want to discount initial payments. Should we add a setting to set whether recurring prices should be affected?

andrewlimaza:
I feel like this should be created as an official Add On that does this, and more. Gists should not be a fully fledged plugin, we should pitch this.

PR #80, opened 2023-02-10
Only discounts regular priced items, excludes any items that are on sale

JarrydLong:
This recipe ensures that only products that are NOT on sale will be discounted when using the Woocommerce Add On. 

All other products will be discounted as expected.

https://gist.github.com/JarrydLong/6fcd0b9a427c96e22ae5b338f8b0b0dd

### DIFF
--- a/add-ons/pmpro-woocommerce/only-discount-regular-price-items.php
+++ b/add-ons/pmpro-woocommerce/only-discount-regular-price-items.php
@@ -19,7 +19,6 @@
 
  function mypmpro_discount_sales( $discount_price, $lowest_price_level, $price, $product ) {
 
-	$regular_price = $product->get_regular_price();
 	$sale_price = $product->get_sale_price();
 
 	if( empty( $sale_price ) ) {
--- a/add-ons/pmpro-woocommerce/only-discount-regular-price-items.php
+++ b/add-ons/pmpro-woocommerce/only-discount-regular-price-items.php
@@ -21,7 +21,7 @@
 
 	$sale_price = $product->get_sale_price();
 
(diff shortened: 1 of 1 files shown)


add-ons/pmpro-woocommerce/only-discount-regular-price-items.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+ function mypmpro_discount_sales( $discount_price, $lowest_price_level, $price, $product ) {
+
+	$regular_price = $product->get_regular_price();
```
dparker1005:
```suggestion
```
`$regular_price` is never used, so we don't need to set it.

add-ons/pmpro-woocommerce/only-discount-regular-price-items.php
```diff
@@ -0,0 +1,34 @@
...
+	$sale_price = $product->get_sale_price();
+
+	if( empty( $sale_price ) ) {
```
dparker1005:
```suggestion
	if (  ! $product->is_on_sale() ) {
```
`is_on_sale()` method checks:
1. That the set sale price is set
2. That the sale price is less than the normal price
3. That a sale is currently running
4. Runs through `woocommerce_product_is_on_sale` filter

https://woocommerce.github.io/code-reference/classes/WC-Product.html#method_get_sale_price

PR #103, opened 2023-06-14
Create load-stripe-library-earlier.php

MaryOJob:
This is an updated version of the previous code to load the Stripe library version earlier restricted to the PMPro checkout page only


JarrydLong (approved):
Looks good to me, if there's any specific use cases that we can follow to test and make 100% let me know

andrewlimaza:
@MaryOJob can you provide more context of this code snippet? Was it used in a ticket - I see the original is quite old (V2.1+)

PR #104, opened 2023-06-15
Rewrite of add Renewal Column

JarrydLong:
This works with the new WP List Table functionality that was implemented in PMPro 2.11


dparker1005:
This is currently the best solution that we have for this functionality, but subs table in 3.0 will make this calculation more accurate. Should we wait for that improved solution to be released, or commit to this code recipe now in the meantime?

dparker1005:
Ah actually I see that this is a rewrite of an existing code recipe, not a new code recipe. The new version is definitely better, merging

PR #105, opened 2023-06-15
Reworks the Discount Code Description

JarrydLong:
This works with the new WP List Table functionality that was implemented in PMPro 2.11


JarrydLong:
Requires this PR in core for the new changes to work - https://github.com/strangerstudios/paid-memberships-pro/pull/2517

PR #107, opened 2023-06-21
Replace Lockdown snippet

kimwhite:
Lock Down Everything But Homepage for Non-Users from https://www.paidmembershipspro.com/lock-down-everything-but-homepage-for-non-users/


JarrydLong (changes requested):
We should redo the formatting so that the brackets and spacing follows latest WPCS standards please

PR #112, opened 2023-08-08
Create set-discount-code-for-use-once-per-user

MaryOJob:


dparker1005:
This code definitely works for now, but in the future, it could be nice to update this logic to a solution that does not rely on user meta and instead checks previous order data directly to determine if a user has used a specific code.

Example here: https://gist.github.com/dparker1005/86f7b7ccf2c79f26a80f7beb40311a81

PR #119, opened 2023-10-13
Restrict all content of a Custom Post Type (CPT) to members only.

JarrydLong:


misc/pmpro-restrict-all-cpt-posts.php
```diff
@@ -0,0 +1,74 @@
...
+    $my_cpt = 'recipe'; // Set your custom post type name here
+
+    // check if post belongs to the CPT
+    if ( is_singular() && get_post_type() === $my_cpt ) {
+        if ( has_action( 'template_redirect', 'pmprocpt_template_redirect' ) ) {
+            remove_action( 'template_redirect', 'pmprocpt_template_redirect' );
+        }
+    }
+}
+add_action( 'wp', 'disable_pmpro_cpt_redirect' );
```
dparker1005, line 74: Do we always want to disable redirects when using this code recipe? Redirects are important for the CPT Add On since our standard restriction code doesn't work for CPTs that aren't filtered by `the_content`. If we disable redirects for these CPTs, then they will always be visible, even to non-members

misc/pmpro-restrict-all-cpt-posts.php
```diff
@@ -0,0 +1,74 @@
...
+function my_pmpro_restrict_all_cpt( $hasaccess, $thepost, $theuser, $post_membership_levels ) {
+
+    // If PMPro says false already, return false.
+    if ( ! $hasaccess ) {
+        // Give admin access
+        if ( current_user_can( 'manage_options' ) ) {
+            return true;
+        } else {
+            return false;
+        }
```
dparker1005, line 26: Do we always want to give admins access as a part of this code recipe? Or should sites use the separate "always give admins access" code recipe if they want that functionality?

andrewlimaza:
I've retested the CPT Add On, and setting "No Redirect" now shows the noaccess message for a CPT. This snippet is no longer needed.

PR #120, opened 2023-10-13
Show a member number on the membership card for Paid Memberships Pro.

JarrydLong:


dparker1005:
Since `generate_member_number()` is a function declared by another code recipe (not included in core PMPro or Add Ons), it may make more sense to include this code with the original Member Number recipe. It looks like this code recipe is the only place that `generate_member_number()` is declared in the snippet library:
https://github.com/strangerstudios/pmpro-snippets-library/blob/1b7724c13775d45574fd0f9772b41100193f7f91/frontend-pages/generate-unique-membership-number.php#L22

Alternatively, maybe we need a Member Number Add On to standardize the functions around this behavior.

PR #121, opened 2023-10-13
Create your own member RSS feed

JarrydLong:


dparker1005:
Looks like this code recipe is currently going into the `misc` folder. Would it make more sense to create an `add-ons/pmpro-member-rss` directory for this to go in?

PR #122, opened 2023-10-13
Updates the level cost text based on the Country selected when using …

JarrydLong:
…the VAT Tax Add On


dparker1005:
Looks good! One note is that a lot of the JS code there is duplicated (once for setup, once for onChange) and can be cleaned up if we wanted to revisit this code in the future, but looks solid

PR #123, opened 2023-10-13
Returns membership profile URL when scanning the QR Code on member's …

JarrydLong:
…Membership Card.

### DIFF
--- a/add-ons/pmpro-membership-card/qr-code-return-membership-profile.php
+++ b/add-ons/pmpro-membership-card/qr-code-return-membership-profile.php
@@ -1,0 +1,36 @@
+<?php
+/*
+ * Returns membership profile URL when scanning the QR Code on member's Membership Card.
+ * 
+ * Learn more at https://www.paidmembershipspro.com/membership-qr-code/
+ *
+ * title: Returns membership profile URL when scanning the QR Code on member's Membership Card.
+ * layout: snippet
+ * collection: add-ons, pmpro-membership-card
+ * category: qr-code
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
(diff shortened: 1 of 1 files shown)


add-ons/pmpro-membership-card/qr-code-return-membership-profile.php
```diff
@@ -0,0 +1,36 @@
...
+ * collection: add-ons, pmpro-membership-card
+ * category: qr-code
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+ function mypmro_membership_card_show_active( $pmpro_membership_card_user, $option ){
```
dparker1005: We should have a more descriptive name here. This function name looks copy/pasted from a different gist.

add-ons/pmpro-membership-card/qr-code-return-membership-profile.php
```diff
@@ -0,0 +1,36 @@
...
+
+	global $pmpro_pages;
+
+	//Checking if PMPro is active
+	if( ! function_exists( 'pmpro_hasMembershipLevel' ) || ! function_exists( 'pmpromd_build_profile_url' ) ){
+		return;
+	}
+
+	if( $option == 'other' ){
+		$profile_url = apply_filters( 'pmpromd_profile_url', get_permalink( $pmpro_pages['profile'] ) );
```
dparker1005: Looks like we don't need to calculate $profile_url here. If a second parameter is not passed to `pmpromd_build_profile_url`, it should build $profile_url in that function.

add-ons/pmpro-membership-card/qr-code-return-membership-profile.php
```diff
@@ -0,0 +1,36 @@
...
+		//Change according to the levels you'd like to check for
```
dparker1005: This comment is also copy/pasted from another recipe

PR #125, opened 2023-10-18
add intval

michaelbeil:
I saw [@dwanjuki's update](https://gist.github.com/dwanjuki/77c47dc61489350a8ebe2d0ef18cf931) to this so I wanted to make sure it was updated


dwanjuki:
For context, using `intval` on line 46 resolves an issue (Ticket #525038) where the same price shows twice for WooCommerce simple products that do not have a member discount. Using loose comparison instead of `intval` worked too. 
pricing
When there's no discount, a `var_dump` of `$price` and `$regular_price` right before line 46 shows that they have the same type and value...I'm not sure why the original code does not work as intended.

PR #131, opened 2023-11-10
Show "Renewal" for orders which are actually renewals

mircobabini:
Snippet to replace "Renewal" label logics showing "Renewal" for orders which are actually renewals.

Orders status column in admin shows a not clear "Renewal" label for returning customers (orders from a customer who already had an order before - might be a renewal or not).

**Some people just want to see "Renewal" for orders which are actually renewals.**

Community request: https://app.slack.com/client/T042RQSG0/C042RQSGW

### DIFF
--- a/orders/renewal-label-for-renewal-orders-only.php
+++ b/orders/renewal-label-for-renewal-orders-only.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * Replace "Renewal" label logics showing "Renewal" for orders which are actually renewals.
+ *
+ * Orders status column in admin shows a not clear "Renewal" label for returning customers
+ * (orders from a customer who already had an order before - might be a renewal or not).
+ *
+ * Some people just want to see "Renewal" for orders which are actually renewals.
+ *
+ * title: Show "Renewal" for orders which are actually renewals.
+ * layout: snippet
+ * collection: orders
+ * category: status, renewal
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+add_action( 'pmpro_orders_extra_cols_header', function ( $order_id ) {
+	echo '<th class="column-status-v2">' . esc_html__( 'Status', 'paid-memberships-pro' ) . '</th>';
+} );
+
+add_action( 'pmpro_orders_extra_cols_body', function ( $order ) {
+	?>
+    <td class="column-status-v2" data-colname="<?php esc_attr_e( 'Status', 'paid-memberships-pro' ); ?>">
+    <span class="pmpro_order-status pmpro_order-status-<?php esc_attr_e( $order->status ); ?>">
+                            <?php esc_html_e( ucwords( $order->status ) ); ?>
+						</span>
+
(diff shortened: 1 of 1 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/pmpro-snippets-library#131 — Show "Renewal" for orders which are actually renewals**
*mircobabini → dev | 1 file, +108 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/131>

---

**Summary**
Good concept, real community need, but not ready to merge. There's a fatal PHP error waiting in `pmpro_order_is_renewal_in_subscription()`, and the snippet hooks into APIs that PMPro core has already flagged for deprecation — which would teach users a pattern that breaks in a future update. Several Copilot suggestions are also correct and should be addressed.

---

**Issues**

**[Critical]** `orders/renewal-label-for-renewal-orders-only.php:78` — `get_original_subscription_order()` returns `false` when no matching order is found (see `class.memberorder.php:897-901`). The snippet then does `$original_subscription_order->id` unconditionally. Any subscription order with no resolvable origin in the DB (migrated data, inconsistency) will throw a fatal "Attempt to read property 'id' on bool." Fix:
```php
$original_subscription_order = $order->get_original_subscription_order();
if ( ! empty( $original_subscription_order ) && $order->id !== $original_subscription_order->id ) {
    $order->is_renewal = true;
} else {
    $order->is_renewal = false;
}
```

**[Major]** `orders/renewal-label-for-renewal-orders-only.php:21,25` — Both `pmpro_orders_extra_cols_header` and `pmpro_orders_extra_cols_body` are marked `// Re-implementing old hook, will be deprecated` in `class-pmpro-orders-list-table.php:162,776`. The snippet adds a duplicate "Status" column and hides the original via CSS — a workaround that will break when PMPro removes these hooks. The newer approach is the `pmpro_order_list_item` filter, which lets you override individual column values directly. This is the bigger design issue and is worth a discussion before finalizing the implementation.

**[Major]** `orders/renewal-label-for-renewal-orders-only.php:83-93` — Non-recurring branch builds SQL via string concatenation with `esc_sql()` instead of `$wpdb->prepare()`. Copilot's suggestion to use `$wpdb->prepare()` with `%d`/`%s` placeholders and typed casts is correct. (Note: the identical query exists in `MemberOrder::is_renewal()` in core — that's not an excuse to repeat the pattern in a snippet library.)

**[Minor]** `orders/renewal-label-for-renewal-orders-only.php:28-29` — `esc_attr_e()` and `esc_html_e()` pass dynamic values through the translation system with no i18n key. These are meant for string literals. Use `echo esc_attr( $order->status )` / `echo esc_html( ucwords( $order->status ) )` instead.

**[Minor]** `orders/renewal-label-for-renewal-orders-only.php:101-107` — CSS is injected via `admin_footer` on every admin page. The body class selector limits visual impact but the `<style>` block fires everywhere. Enqueue on the orders screen only via `get_current_screen()` inside a `current_screen` action.

**[Minor]** `orders/renewal-label-for-renewal-orders-only.php:33-38` — The "Renewal" link always passes `$order->subscription_transaction_id` as the search param `s`, but for non-recurring renewals (the `else` branch) this field is empty, producing a useless link. Either omit the link for non-recurring renewals or render a plain `<span>` instead (Copilot's suggestion is reasonable here).

---

**Questions**

1. Is the intent to replace the existing `$order->is_renewal()` method behavior entirely, or only supplement it for the orders list column? If the former, the cleaner path is a filter on `is_renewal` rather than a parallel function.
2. Has the "Needs Thought" label been resolved? The label is still on the PR.

PR #133, opened 2023-12-19
Disable the PMPro redirect from default WordPress register page to PM…

MaximilianoRicoTabo:
…Pro levels page.


andrewlimaza:
Since this is a one-liner snippet I don't think we'd add it into the library :) These are okay to stay as gists.

PR #150, opened 2024-04-11
Tightening up the recipe to try to get more reliable data in the MMPU…

kimcoleman:
… context.


kimcoleman:
Merging for now - we are also (in core PMPro) going to just do this by default. Once that is done we will remove this recipe method. https://github.com/strangerstudios/paid-memberships-pro/pull/2954

PR #152, opened 2024-04-13
Mega code recipe for limiting members in a level.

kimcoleman:
This is a big gist and could probably be a plugin some day. But its a good example of end to end how you add a setting and work with that setting.

This is 3.0 compatible - where the old gists like these were not:
- https://gist.github.com/kimcoleman/51c03c3ec77ab84713dc80a512225f11
- https://gist.github.com/andrewlimaza/79f29484abd7713cf4de0e2cd6cdedae

If this gets approved we can update this blog post and consolidate it: https://www.paidmembershipspro.com/limit-number-members-membership-level/

There are not settings once the gist is added to the site. They will just need to edit their Membership Level settings.

Note: This recipe has a different approach than the linked old gists - which used an option, not level meta, to store level limits. The code is not backwards compatible with the old gists.

### DIFF
--- a/membership-levels/limit-members-level.php
+++ b/membership-levels/limit-members-level.php
@@ -57,23 +57,16 @@ add_action( 'pmpro_membership_level_after_other_settings', 'my_pmpro_limit_membe
 function my_pmpro_limit_members_pmpro_registration_checks( $value ) {
 	global $wpdb;
 	
-	// Default to level passed in via URL.
-	if ( empty( $level_id ) && ! empty( $_REQUEST['pmpro_level'] ) ) {
-		$level_id = intval( $_REQUEST['pmpro_level'] );
-	}
+	// Get the level at checkout.
+	$pmpro_level = pmpro_getLevelAtCheckout();
 
-	// If we don't have a level, check the legacy 'level' request parameter.
-	if ( empty( $level_id ) && ! empty( $_REQUEST['level'] ) ) {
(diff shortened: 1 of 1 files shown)


membership-levels/limit-members-level.php
```diff
@@ -0,0 +1,249 @@
...
+function my_pmpro_limit_members_pmpro_membership_level_after_other_settings() {
+	$edit_level_id = $_REQUEST['edit'];
+	$pmpro_level_member_limit = get_pmpro_membership_level_meta( $edit_level_id, 'pmpro_level_member_limit', true );
+	?>
+	<table class="form-table">
+		<tbody>
+			<tr>
+				<th scope="row" valign="top"><label for="pmpro_level_member_limit"><?php esc_html_e( 'Maximum Number of Members', 'pmpro-customizations' ); ?></label></th>
+				<td>
+					<input id="pmpro_level_member_limit" name="pmpro_level_member_limit" type="number" value="<?php echo esc_attr( $pmpro_level_member_limit );?>" />
```
MaximilianoRicoTabo, line 44: we could leverage type number here and add "min" attribute to avoid negative numbers

MaximilianoRicoTabo (approved):
LGTM

andrewlimaza:
Something to think about is to handle renewals, but I'm going to merge it in for now. It definitely can become an Add On in the near future.

PR #156, opened 2024-05-08
Create a discount code for expired members

JarrydLong:
To be used on https://www.paidmembershipspro.com/discount-code-expired-members/


dparker1005:
Is the intention here that only the user the code is created for can use the code? If so, it would be good to add a registration check to make sure that other users can't check out with the code (ex guess codes `PMPRO1`, `PMPRO2`, `PMPRO3`, etc)

PR #166, opened 2024-06-11
Disable All PMPro Email

kimwhite:
Disable All PMPro Email from post: https://www.paidmembershipspro.com/disable-emails-member-user-admin/

### DIFF
--- a/email/disable-all-pmpro-email.php
+++ b/email/disable-all-pmpro-email.php
@@ -1,0 +1,26 @@
+<?php
+/**
+* Automatically Disable All PMPro Email
+* The function below will disable all email sent by Paid Memberships Pro –
+* - both for the member and the administrator.
+*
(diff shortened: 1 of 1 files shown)


email/disable-all-pmpro-email.php
```diff
@@ -0,0 +1,26 @@
...
+* https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+*/
+
+
+function my_pmpro_disable_all_email($recipient, $email)
+{
+    //if($email->template == "invoice")   //use this to check for a certain template
+    $recipient = NULL;
+    return $recipient;
+}
```
dparker1005:
```suggestion
function my_pmpro_disable_all_email( $recipient, $email ) {
    return null;
}
```

PR #188, opened 2024-08-02
* Add Format the phone field to look like (###) ###-#### recipe

MaximilianoRicoTabo:
image

### DIFF
--- a/checkout/format_phone_field.php
+++ b/checkout/format_phone_field.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Format a custom field to collect the phone field to look like (###) ###-####
+ *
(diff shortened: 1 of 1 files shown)


andrewlimaza:
@MaximilianoRicoTabo here is the guide we currently have that is similar to this one - https://www.paidmembershipspro.com/how-to-format-the-member-phone-number-field-using-a-filter/

If you have a link to the original recipe, please let me know so I can take a look at it further. This seems to be a new one/way of doing the format phone number field.

MaximilianoRicoTabo:
Good to know. This PR comes from here https://paidmembershipspro.slack.com/archives/G04GXBGFL/p1722605613634249 I think.

PR #216, opened 2024-11-15
Wp Bouncer Pop Up

kimwhite:
wp-bouncer
 * Allows for a popup to show user has exceeded their log in limit 
 * and then gives them the option to log out of other instances using jquery and ajax.


andrewlimaza:
This needs to be redone for the Limit Login Add On instead.

andrewlimaza:
We are working on a different snippet as it's no longer available on our site. It should still work as we have backwards compatibility for WP Bouncer to Limit Login Add On.

PR #219, opened 2024-11-15
New Login-Logout Folder and 2 restrict login recipes

kimwhite:
new folder, for login logout recipes.
Added 2 recipes from this post:
https://www.paidmembershipspro.com/restrict-user-login-for-members-only/


andrewlimaza:
Moved this back to `misc`, it's okay to keep it in `misc` and then set the category to `login`

PR #225, opened 2024-11-18
Create Create A Unique Email For Membership Renewals in PMPro

MaryOJob:


andrewlimaza (changes requested):
Requires WPCS and correct docblock layout.  A nice to have might be a fallback if the file doesn't exist in the email directory to use default text so it doesn't break completely or send empty emails.

Please can you make these adjustments.

MaryOJob:
@andrewlimaza I updated this snippet, wanted to ask too; should we still be adding custom email templates like this or this should be done using the `pmproet_templates` filter?

andrewlimaza:
Yes it would be best to add it to the `pmproet_templates` filter and the linked documentation would need to be updated (once merged in).

MaximilianoRicoTabo:
Hey @andrewlimaza @MaryOJob what do you think about hooking into
`pmpro_send_checkout_emails` filter https://github.com/strangerstudios/paid-memberships-pro/blob/v3.4/includes/checkout.php#L299

and then instead of send the checkout email instantiate a new custom renewal class created as a child of PMPro_Email_Template  class and send it. A more v3.4 approach. I can test further tomorrow if you think it worths

andrewlimaza:
We are bringing out liquid templates for more conditional logic. This is no longer needed.

PR #229, opened 2024-11-25
Create Add A Percentage Based Fee to Checkout When Using Stripe

MaryOJob:


MaximilianoRicoTabo:
this recipe could use pmpro_getGateway() instead of taking from the Request object. Other than that it looks good to me @andrewlimaza 

EDIT: Doesn't work with the gateway in the request. I'll open a new one

andrewlimaza:
This has been merged in.

PR #248, opened 2025-01-24
Turn WooStore into a Catalog

kimwhite:
Non Members will not be able to purchase.

https://www.paidmembershipspro.com/turn-your-woocommerce-store-into-a-catalog-for-non-members/


kimwhite:
already done

PR #249, opened 2025-01-24
Update the Level Cost Text Based on Country and State Upldate cost with tax

kimwhite:
https://www.paidmembershipspro.com/update-level-cost-text-country-state/

### DIFF
--- a/checkout/update-level-cost-with-tax.php
+++ b/checkout/update-level-cost-with-tax.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Custom tax structure where all levels except level 1 have 7.25% tax if billing state is CA.
+ *
+ * title: Custom tax structure where all levels except level 1 have 7.25% tax if billing state is CA.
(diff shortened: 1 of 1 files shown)


checkout/update-level-cost-with-tax.php
```diff
@@ -0,0 +1,70 @@
...
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+ function mypmpro_tax_update_script(){
+
+	global $pmpro_levels;
+
+	$level = ( isset( $_REQUEST['level'] ) ? $pmpro_levels[$_REQUEST['level']] : '' );
```
andrewlimaza, line 21: This needs to support ?pmpro_level, we can use pmpro_getLevelAtCheckout() instead of getting it like this from the global.

PR #250, opened 2025-01-24
Show a Different “Terms of Service” at Checkout Based on Membership Level

kimwhite:
Updated from original to remove warning messaage on settings page.
(https://gist.github.com/strangerstudios/9073376#file-my_option_pmpro_tospage-php

for this post
https://www.paidmembershipspro.com/show-different-terms-service-checkout-based-membership-level/


andrewlimaza (changes requested):
This needs to run through WPCS.

PR #270, opened 2025-03-03
Update member-count-shortcode.php

MaryOJob:
This new recipe now shows the correct member count per level, and it would also now work to show two levels within the shortcode.


andrewlimaza:
@MaryOJob I tried to simplify the conditionals when outputting data, please review the adjustments when you have a chance.

PR #271, opened 2025-03-07
Hides multiple BP profile fields from Profile View until user is approved.

kimwhite:
BuddyPress has a setting on it's Profile Fields for visibility. There is a setting for Member Only, which really means user only. 
When a site is using the Approval Add On, this snippet will continue to hide the fields until the PMPro Memberships is approved.
Screenshot 2025-03-07 at 8 04 41 AM


andrewlimaza:
Thanks for this, I reworked the logic to not allow non-members or logged-out members a way to view these fields. The snippet is more 'generalized' now and developers can tweak it to target specific membership levels instead should they have fields per level ID or role etc and need to tweak it even further.

The `isApproved` method will return true if the member has a level that doesn't require approval so it's okay to leverage that method's logic without checking if the level requires approval.

PR #276, opened 2025-03-17
Create - Add Account page to Woo My Account Navigation

kimwhite:
Using WooCommerce Endpoints, add the PMPro Membership Account shortcode to the My Account page


kimwhite:
need to update

PR #279, opened 2025-03-18
Login, hide username and email fields in error message

kimwhite:
For greater security, don't show username and email information in the login error message.
Can be customized to create your own message, removing the %s information.

Created folder for all Login gist (and logout)


andrewlimaza:
I'm good with this change, I did add a feature to make the login form response messages on PMPro filterable and if approved we can adjust this snippet - merging. Thanks!

I did move things back to their original state, we use categories and tags to try and organize structure and would like to avoid subfolders for now (until we have a generated site from the repo and know how we want the file structure).

PR #285, opened 2025-04-03
Display elements to membership levels based on element class for Bricks Builder

MaryOJob:


andrewlimaza:
We have a feature to go out in a future version for Bricks integration -> https://github.com/strangerstudios/paid-memberships-pro/pull/3342

andrewlimaza:
This is another way of showing/hiding content based on CSS class which is great and offers more flexibility :)

PR #293, opened 2025-04-11
Create shortcode to show all PMPro courses with lesson count

MaryOJob:
Create a custom shortcode to show all PMPro courses with their lesson count in a tabularized layout.


show-all-courses-with-lesson-count.php
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        echo '</div>';
+
+        return ob_get_clean();
+    }
+
+    add_shortcode( 'pmpro_courses_with_lessons', 'pmpro_courses_with_lessons_shortcode' );
+}
+add_action( 'plugins_loaded', 'pmpro_courses_with_lessons_init' );
```
andrewlimaza: Is this needed for creating the custom shortcode? This should be able to ensure the functions exist and merged into just the shortcode functionality.

We can use `function_exists` to ensure the Courses Add On is installed. I also wonder if this is better suited to be submitted to PMPro Courses instead (as an attribute to existing shortcode or something similar).

andrewlimaza (changes requested):

show-all-courses-with-lesson-count.php
andrewlimaza: I have made these changes.

MaryOJob: Got it, Andrew. I will go through this 💃🏽

PR #304, opened 2025-05-06
Create my-pmpro-membership-card-create-display-custom-field.php

davidmutero:


andrewlimaza:
Adding a note here, we need to refactor this to use `PMPro_Field` class instead of the Register Helper class.

PR #308, opened 2025-05-07
Create Add Widget to Filter Members Based On Location and Distance

MaryOJob:


andrewlimaza:
This has been rewritten to support the new way of Membership Maps storing data.

PR #311, opened 2025-05-20
Update remove-element-from-default-dimensions-array.php

MaryOJob:
The existing snippet was throwing a `Warning: Undefined variable $dimension` error, which resulted from a typo on line 25. This PR update resolves this error.


andrewlimaza:
Nice catch Mary, thanks for this!

PR #325, opened 2025-07-01
Update the strikethrough pricing code to work for variable products

MaryOJob:
- Replaced short array syntax [] with array() for WPCS compatibility.
- Added multiline formatting to long conditional statements for readability.
- Used instanceof WC_Product check before using variation objects.
- Added empty() checks before using variables like $regular_price and $level_id.
- Reformatted comments to follow WPCS standards (capitalization, punctuation).
- Removed shorthand ternary logic in favor of clearer if structure where needed.
- Ensured consistent spacing around operators and function arguments.
- Split long lines and nested function calls across multiple lines.
- Added fallback checks to avoid undefined properties (e.g., $current_user->membership_level->id).

Screenshot 2025-07-01 at 3 37 44 pm

show-strikethrough-pricing-for-members-only-in-woo-store

### DIFF
--- a/add-ons/pmpro-woocommerce/strike-through-pricing-woocommerce.php
+++ b/add-ons/pmpro-woocommerce/strike-through-pricing-woocommerce.php
@@ -13,85 +13,107 @@
  * Read this companion article for step-by-step directions on either method.
  * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
  */
-
 function my_pmprowoo_strike_prices( $price, $product ) {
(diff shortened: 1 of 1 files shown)


andrewlimaza (approved):

andrewlimaza:
We've made changes that if the variation product has a single amount (and not a ranged amount) it shows the single value. For example instead of showing $10.00 - $10.00 it will just show $10.00 as this is how WooCommerce handles it as well in core.

Various optimizations have been added to this code snippet to make it more stable and even closer to getting it to be able to be merged into PMPro WooCommerce with minimal effort.

PR #335, opened 2025-07-30
Add Link, Style Message

kimwhite:
Added link to the post with this recipe.

Added style to make the Message to save first easier to see.


kimwhite:
Examples
Default
Description does not show up until code is saved

With added style
with styling

kimwhite:
Updated using style

message

PR #347, opened 2025-08-26
Preserve membership time into recurring renewal

kimwhite:
When a member with a **one-time or fixed-term membership** renews into a r**ecurring subscription** for the **same level**, skip the initial payment and set the first recurring payment to begin when their current expiration date ends.


andrewlimaza:
This needs that small adjustment to the level_cost_text.

andrewlimaza (changes requested):
This needs to be updated as per our discussion.

PR #349, opened 2025-09-02
Create custom user avatars with user fields

MaryOJob:
This was previously created in the Add Ons folder in the Snippets Library under Register Helper.

This snippet is an updated version added to the User Fields folder using the User Fields functionality to create Custom Avatars.

### DIFF
--- a/user-fields/custom-user-avatars-with-user-fields.php
+++ b/user-fields/custom-user-avatars-with-user-fields.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * Create Custom User Avatars with the User Fields
+ *
+ * Allow members to upload their avatar using a User field during checkout or on the Member Profile Edit page.
+ *  
+ * title: Create Custom User Avatars with User Fields
+ * layout: snippet
+ * collection: user-fields
+ * category: user-avatars
+ * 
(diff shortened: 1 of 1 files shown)


user-fields/custom-user-avatars-with-user-fields.php
```diff
@@ -0,0 +1,83 @@
...
+            'label'         => 'My Photo',               // custom field label
+            'hint'          => 'Recommended size is 100px X 100px',
+            'profile'       => true,                   // show in user profile
+            'preview'       => true,                   // show a preview-sized version of the image
+            'allow_delete'  => true,                   // Allow user to see Delete option on Avatar
+        )
+    );
+
+    // Add a field group to put our fields into
+    pmpro_add_field_group( 'My Profile Picture' );
```
dparker1005: When calling `pmpro_add_field_group()`, we should try to start passing two parameters:
1. `name` - This is a computer-friendly name usually without spaces. For example "my-profile-picture"
2. `label` - This is the text to appear at the top of the group at checkout. For example "My Profile Picture"

And the `pmpro_add_user_field()` call below will need to be updated to use the new `name` as well

dparker1005 (changes requested):

PR #375, opened 2025-10-02
Create style-locked-message-overlay-on-protected-content.css

MaryOJob:
Adding this CSS to the snippets library.


andrewlimaza:
CSS snippets should be left on gist.github.com for now. One thing I noticed is that the `..pmpro_protected_post_blur_mask` should have a `z-index:1` or higher value to correctly work.

MaryOJob:
> CSS snippets should be left on gist.github.com for now. One thing I noticed is that the `..pmpro_protected_post_blur_mask` should have a `z-index:1` or higher value to correctly work.

This has been implemented.

PR #376, opened 2025-10-03
Create override-default-map-start-location.php

MaryOJob:
Adding to Snippets Library with updated filter and minor WPCS fixes


andrewlimaza:
I merged in the previous PR that was unrelated and no longer needed to be correct. See https://github.com/strangerstudios/pmpro-snippets-library/blob/dev/add-ons/pmpro-member-directory/override-default-map-start-location.php

PR #387, opened 2025-10-15
Early Renewal Discount

kimwhite:
snippet for this post
http://gist.github.com/strangerstudios/d0b06dbb36483a76b030#file-erd_pmpro_checkout_level-php


kimwhite:
Here is what it looks like with the new function
early renwal message

PR #407, opened 2025-12-09
Add Process Membership Checkout in a Modal or Popover Window

dwanjuki:
Migrating snippet to library: https://gist.github.com/ideadude/d25fb89822f60143926071feace36812

Updated `pmprocm_content()` to check if `$queried_object` and `post_content` property is empty, resolves warnings:
```
strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated
Attempt to read property "post_content" on null
```
Also checking for existence of the Checkout Page block here. The original only checked for the pmpro_checkout shortcode. 

Some WPCS changes and added recipe header.


andrewlimaza:
Thanks for doing this one. 👍

PR #419, opened 2026-01-13
Change who get's PMPro admin emails

kimwhite:
by default the site admin gets these, with this recipe you can Change the default And set BCC if you need too.


kimwhite:
from this original
https://gist.github.com/andrewlimaza/78a4e1fb266b633aa5b138047576bfcc

andrewlimaza:
Removed BCC function as we have it here already -> https://github.com/strangerstudios/pmpro-snippets-library/blob/dev/email/bcc-email-addresses-on-notifications.php

PR #439, opened 2026-04-10
Discount code show price change on checkout page when applied

kimwhite:
This is an old snippet to show the updated price and the old price when using a discount code.
Original Here https://gist.github.com/messica/33c769db1267f7aaf3741f1cf0989de2#file-my_pmpro_applydiscountcode_return_js-php

I added a line to show it above the submit button, also.
Also shows recurring billing price if changed.
show pricing

show pricing bottom


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#439 — Discount code show price change on checkout page when applied**
*kimwhite → dev | 1 file, +79 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/439>

---

**Summary**

New snippet that hooks `pmpro_applydiscountcode_return_js` to display original price, discounted price, savings, and updated recurring billing when a discount code is applied. The approach is sound and the logic is correct. Not ready to merge as-is — file has a double extension and the frontmatter has copy-paste errors.

---

**Issues**

- **Major** `discount-codes/display-discount-savings-at-checkout.php.php` — Double `.php.php` extension on the filename. Every other file in `discount-codes/` uses a single `.php`. Rename to `display-discount-savings-at-checkout.php`.

- **Minor** `display-discount-savings-at-checkout.php.php:7` — `title: Add description to discount codes` is copy-pasted from another snippet and doesn't describe this one. Should be something like `title: Display original and discounted price when a discount code is applied`.

- **Minor** `display-discount-savings-at-checkout.php.php:9` — `category: description` doesn't match. The `add-description-to-checkout-field-for-admin.php` snippet uses that category. This snippet's category should reflect pricing or checkout display (check what other discount-code snippets use in this library).

- **Minor** `display-discount-savings-at-checkout.php.php:57` — `$cycle_period` is interpolated directly into `$html` without escaping: `"Then {$discounted_recurring_price} per {$cycle_number} {$cycle_period}"`. In practice `cycle_period` is always a PMPro enum value (Day/Week/Month/Year), but snippets library code gets copied widely — use `esc_html( $level->cycle_period )` here.

- **Minor** `display-discount-savings-at-checkout.php.php:51` — `$original_recurring != $discounted_recurring` compares floats with `!=`. Use `abs( $original_recurring - $discounted_recurring ) > 0.001` to avoid false positives from floating-point representation.

- **Minor** `display-discount-savings-at-checkout.php.php:79` — Missing newline at end of file.

---

**Looks Good**

- The `wp_json_encode()` wrapping on the jQuery append calls is the right approach — prevents HTML in `$html` from breaking the JS output.
- The guard at line 18 (`empty( $code_level )`) and the `pmpro_getLevel()` null check correctly bail before any work happens.
- Clearing stale messages with `jQuery('.pmpro-discorig-message').remove()` before re-appending prevents duplicates on repeated code entry.

---

**Questions**

1. Does `#pmpro_message_bottom` exist in all PMPro checkout templates, or only the default Elementor/block checkout? Worth a note in the snippet comments if it's template-dependent.

PR #446, opened 2026-04-27
Creditcard image checkout

kimwhite:
Update these to work with New PayPal Gateway Add On

### DIFF
--- a/checkout/add-cc-and-paypal-logos-checkout-page.css
+++ b/checkout/add-cc-and-paypal-logos-checkout-page.css
@@ -25,7 +25,6 @@
 
 .pmpro-checkout .gateway_authorizenet .pmpro_form_label,
 .pmpro-checkout .gateway_braintree .pmpro_form_label,
-.pmpro-checkout .gateway_paypal .pmpro_form_label,
 .pmpro-checkout .gateway_stripe  .pmpro_form_label {
 	background-image: url(images/pay-with-credit-cards.png);
 	background-position: left bottom;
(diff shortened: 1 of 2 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/pmpro-snippets-library#446 — Creditcard image checkout**
*kimwhite → dev | 2 files, +3 -3 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/446>

**Summary**
The CSS change is correct. The PHP change introduces a broken image URL and needs to be reverted.

**Issues**

- **Critical** `checkout/add-cc-logos-checkout-page.php:32` — Replacing `plugins_url()` with `get_site_url()` breaks the image for everyone.

  Before: `plugins_url( '/images/pay-with-credit-cards.png', __FILE__ )` → correctly resolves to the snippet's own directory, e.g. `https://example.com/wp-content/mu-plugins/images/pay-with-credit-cards.png`

  After: `get_site_url( null, '/images/pay-with-credit-cards.png' )` → resolves to the site root, e.g. `https://example.com/images/pay-with-credit-cards.png`

  There's no image file at the site root. This will be a broken `` on every install. Revert to the original `plugins_url( '/images/pay-with-credit-cards.png', __FILE__ )`.

  This change also has nothing to do with the New PayPal Gateway — it looks like an unrelated edit that shouldn't be in this PR.

**Looks Good**

The CSS change in `add-cc-and-paypal-logos-checkout-page.css` is the right fix. Moving `.gateway_paypal .pmpro_form_label` from the credit-card selector group to sit alongside `.gateway_paypalexpress` in the PayPal selector correctly aligns the logo display with how the new PayPal gateway class works.

PR #450, opened 2026-05-07
Fix one-time CC required trial snippet: level ID mismatch, naming, hook type

kimcoleman:
Summary

- Fix copy-paste bug where $trial_level_id = 9 in the checkout-side function didn't match $trial_level_id = 1 in the save-side function, leaving the delay-removal path as dead code on the example level
- Rename one_time_trial_delay_pmpro_registration_checks() to my_pmpro_one_time_trial_remove_delay_hooks() so it follows the my_pmpro_ prefix used by every other function in the snippet, and so the name reflects what it actually does (it removes Subscription Delays Add On hooks at init; it does not run on pmpro_registration_checks) 
- Change add_filter( 'init', ... ) to add_action( 'init', ... ) since init is an action

Notes  

- remove_filter / remove_action targets verified against pmpro-subscription-delays v0.6.1 — all callback names and priorities match
- Priority 5 on pmpro_checkout_level confirmed to run before the Add On's default priority 10 callback 

Test plan

- First checkout for the trial level applies the configured subscription delay 
- Second checkout for the same level by the same user charges the billing amount immediately (no delay)
- Edit Profile page shows "Trial period has been claimed." after first checkout
- Levels page shows the recurring price (not the trial price) for users who already claimed the trial  
- Discount-coded checkouts still respect the Add On's delay (snippet bails early when a discount code is present)


flintfromthebasement (approved):
**PR: strangerstudios/pmpro-snippets-library#450 — Fix one-time CC required trial snippet: level ID mismatch, naming, hook type**
*kimcoleman → dev | 1 file, +6 -5*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/450>

**Summary**
Three real bugs fixed, all correctly. The level ID mismatch was the most damaging — `$trial_level_id = 9` in the remove-delay function made the entire repeat-customer path dead code for sites using level 1 (the stated example). The hook type fix and rename are both correct. Ready to merge.

**Looks Good**
- The PR description is thorough: level ID rationale, naming rationale, `add_filter` → `add_action` rationale, and explicit notes that the `remove_filter`/`remove_action` targets were verified against pmpro-subscription-delays v0.6.1. That's the right way to document a snippet fix.
- Priority 5 on `pmpro_checkout_level` (line 132) correctly runs before the Add On's default priority 10, so the snippet's level mutation lands before the Add On would have applied the delay. The interaction is sound.
- Trailing newline added. Housekeeping done.

**Questions**
One pre-existing issue worth noting for a follow-up, not blocking this PR: `my_pmpro_one_time_trial_show_trial_level_used()` around line 44 checks `$already == '1'` (hardcoded) instead of comparing to `$trial_level_id`. If an admin customizes the trial level to anything other than 1, the "Trial period has been claimed" display on the Edit Profile page will always show "not claimed" regardless of actual state. Low-stakes (admin-only display), but worth fixing in a separate pass since it undermines the snippet's usefulness as a configurable example.

PR #455, opened 2026-05-11
Add affiliate cookie assignment from utm_source

MaryOJob:
This snippet auto-assigns a PMPro affiliate cookie based on the incoming utm_source if a matching affiliate code exists.

### DIFF
--- a/add-ons/pmpro-affiliates/set-affiliate-cookie-from-utm-source.php
+++ b/add-ons/pmpro-affiliates/set-affiliate-cookie-from-utm-source.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * Auto-assign a PMPro affiliate cookie based on incoming utm_source.
+ * Only sets the cookie if a matching, enabled affiliate code exists.
+ *
+ * title: Auto-assign affiliate cookie from utm_source
+ * layout: snippet
+ * collection: add-ons
+ * category: pmpro-affiliates
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#455 — Add affiliate cookie assignment from utm_source**
*MaryOJob → dev | 1 file, +35 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/455>

**Summary**
Single new snippet that fires on `init`, checks for `utm_source` in the query string, looks up a matching enabled affiliate code in the DB, and sets the `pmpro_affiliate` cookie if found. Sanitization and prepared statement are correct. Two minor issues to fix before merge.

**Issues**

- **Minor** `set-affiliate-cookie-from-utm-source.php:24-29` — The snippet queries `{$wpdb->prefix}pmpro_affiliates` directly with no check that the PMPro Affiliates plugin is active. If the plugin is inactive the table won't exist; `$wpdb->get_row()` will silently return `null` on a failed query, but this is an inconsistency with the library's own pattern. Every sibling snippet in this directory guards with a `function_exists()` call first (see `create-affiliates-bulk.php:27`). Add before the DB call:
  ```php
  if ( ! function_exists( 'pmpro_affiliates_getNewCode' ) ) {
      return;
  }
  ```

- **Minor** `set-affiliate-cookie-from-utm-source.php:9` — `link: TBD` is a placeholder that will break the snippets site renderer. Needs a real URL or should be removed before merge.

**Looks Good**

`sanitize_text_field( wp_unslash( $_GET['utm_source'] ) )` before the prepared statement is correct. The early-return on existing cookie (`! empty( $_COOKIE['pmpro_affiliate'] )`) correctly avoids stomping an already-set affiliate attribution.

PR #456, opened 2026-05-14
Add require-approval-for-every-signup.php

davidmutero:
### What This Does
Forces all signups for approval-required membership levels to remain in pending approval status after checkout.

### Why It's Helpful
Useful for sites that require manual review and approval of every new member signup before granting access.

### Technical Notes
- Uses the `pmpro_after_checkout` action.
- Checks whether the level requires approval using the PMPro Approvals Add On.
- Updates the user's approval metadata to pending.
- Clears the approval count transient cache.

### Dependencies
Requires the PMPro Approvals Add On.

### DIFF
--- a/add-ons/pmpro-approvals/require-approval-for-every-signup.php
+++ b/add-ons/pmpro-approvals/require-approval-for-every-signup.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Require Approval for Every Signup
+ * 
+ * This recipe forces every checkout to require approval.
+ *
+ * Requires the Approvals Add On to be active for this to work.
+ *
+ * title: Require Approval for Every Signup
+ * layout: snippet
+ * collection: add-ons, pmpro-approvals
+ * category: approvals
+ * link: TBD
+ * 
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+function my_pmpro_require_approval_for_every_signup( $user_id, $morder ) {
+
+	if ( ! class_exists( 'PMPro_Approvals' ) ) {
+		return;
+	}
+
+	$level_id = $morder->membership_level->id;
+
+	if ( ! PMPro_Approvals::requiresApproval( $level_id ) ) {
+		return;
+	}
+
+	update_user_meta(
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#456 — Add require-approval-for-every-signup.php**
*davidmutero → dev | 1 file, +47 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/456>

---

**Summary**

The snippet's real job is forcing previously-approved members back to `pending` on every new checkout, which the Add On's own `pmpro_before_change_membership_level` skips for already-approved users. That's a legitimate gap to fill. The hook choice and priority 99 are correct — level changes complete before `pmpro_after_checkout` fires. Two issues need fixing before merge: wrong timestamp function and a missing audit log update.

---

**Issues**

**Major** `require-approval-for-every-signup.php:37` — `time()` produces a UTC Unix timestamp, but `updateUserLog()` displays it via `date_i18n( get_option('date_format'), $timestamp )`, which applies the WordPress site timezone to the value. The rest of the Add On uses `current_time('timestamp')` for this field (see `approveMember` and `resetMember`), which returns local time as a Unix integer. Using `time()` here will show approval timestamps offset by the site's UTC offset in the admin UI on non-UTC sites. Fix: replace `time()` with `current_time('timestamp')`.

**Major** `require-approval-for-every-signup.php:32-43` — Every status transition in the Add On (`approveMember`, `denyMember`, `resetMember`) calls `PMPro_Approvals::updateUserLog( $user_id, $level_id )` after writing the meta. This snippet skips that call, so the forced pending reset is invisible in the admin approval history. Admins reviewing the approval log will see no record of why the user is back in pending. Fix: add `PMPro_Approvals::updateUserLog( $user_id, $level_id );` after the `delete_transient` call.

**Minor** `require-approval-for-every-signup.php:7` — The file header says "Forces every checkout to require approval" but the code only targets levels already configured with `requiresApproval() === true`. Levels without the approval checkbox ticked are skipped. The comment should say something like "Forces re-signups for approval-required levels back to pending, even if the user was previously approved."

**Minor** `require-approval-for-every-signup.php:47` — Missing newline at end of file (`\ No newline at end of file` in the diff).

---

**Looks Good**

Checking `class_exists('PMPro_Approvals')` before doing anything is the right defensive pattern for a dependency-gated snippet. Priority 99 on `pmpro_after_checkout` correctly places this after the Add On's own `pmpro_before_change_membership_level` logic, so the override lands last.

PR #458, opened 2026-05-17
Adding snippet for a coming soon redirect as well as snippets for hardening built off our internal team WP site's logic

kimcoleman:
Snippet for a coming soon page redirect for non-admins.

Plus, four snippets for hardening WordPress, focused on four user-leak areas for

- /?author=N redirect
- /wp-sitemap-users-1.xml
- /wp-json/wp/v2/users
- Login error messages

Plus misc/rate-limit-password-resets.php for brute-force mitigation. That's a clean set for the hardening post to anchor around.

### DIFF
--- a/misc/block-author-enumeration.php
+++ b/misc/block-author-enumeration.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Block author enumeration via ?author=N requests.
+ *
+ * title: Block author enumeration to hide usernames
+ * layout: snippet
+ * collection: misc
+ * category: security, login
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+/**
+ * By default, WordPress will resolve a URL like /?author=1 to the author's
+ * archive, which exposes the user's login slug in the redirect URL. This
+ * snippet redirects those requests to the home page for logged-out users
+ * before WordPress resolves the slug.
+ */
+function my_block_author_enumeration() {
+	if ( ! is_user_logged_in() && isset( $_GET['author'] ) ) {
+		wp_safe_redirect( home_url() );
+		exit;
+	}
+}
+add_action( 'init', 'my_block_author_enumeration' );
--- a/misc/block-author-enumeration.php
+++ b/misc/block-author-enumeration.php
@@ -17,11 +17,15 @@
 /**
  * By default, WordPress will resolve a URL like /?author=1 to the author's
  * archive, which exposes the user's login slug in the redirect URL. This
- * snippet redirects those requests to the home page for logged-out users
- * before WordPress resolves the slug.
+ * snippet redirects those requests to the home page for anyone without
+ * the edit_posts capability — so admins, editors, and authors can still
+ * view author archives while subscribers and logged-out visitors can't.
+ *
+ * If you want author enumeration available to all logged-in users (subscribers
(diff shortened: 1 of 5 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#458 — Adding snippet for a coming soon redirect as well as snippets for hardening**
*kimcoleman → dev | 5 files, +184 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/458>

---

**Summary**

Five new snippets: a coming-soon redirect for non-admins, plus a security hardening set covering author enumeration, REST API user exposure, the user sitemap, and password-reset rate limiting. The approach is solid and the snippets are well-commented. Two issues in the rate-limiter are worth addressing before this gets published as canonical advice.

---

**Issues**

**Major** `misc/rate-limit-password-resets.php:17` — `$_SERVER['REMOTE_ADDR']` is the proxy IP on sites behind Cloudflare or any load balancer. In that case all visitors share a single rate-limit bucket, so 3 password reset attempts from anyone — including a legitimate user — blocks the entire site from password resets for an hour. This is a latent DoS on a core auth flow. Add a note in the docblock: "If your site is behind Cloudflare, use `$_SERVER['HTTP_CF_CONNECTING_IP']` instead; if behind a generic proxy, use `HTTP_X_FORWARDED_FOR` with validation. Trusting these headers directly on a non-proxied site allows IP spoofing."

**Major** `misc/rate-limit-password-resets.php:20-30` — Classic transient race condition. Two concurrent requests from the same IP can both read `$count = 0` before either writes, passing the `>= 3` check simultaneously. For a published security snippet, add a comment acknowledging this: transient rate-limiting is best-effort and should be paired with server-level throttling (Nginx `limit_req`, Cloudflare rate rules) for hard enforcement.

**Minor** `restricting-content/redirect-non-admins-to-coming-soon-page.php:38-39` — `strpos( $_SERVER['REQUEST_URI'], 'wp-admin' )` matches any URL containing that substring, not just the admin directory. A request to `/wp-administrators/` or `/?source=wp-admin-docs` bypasses the redirect. Change to:
```php
strpos( $_SERVER['REQUEST_URI'], '/wp-admin/' ) === 0
    || strpos( $_SERVER['REQUEST_URI'], '/wp-login.php' ) === 0
```

**Minor** `misc/block-rest-api-users-endpoint.php:39` — Returns HTTP 401 (Unauthorized), which implies the client should authenticate to gain access. Authenticated users without `list_users` capability also hit this path and get a misleading 401. Use 403 (Forbidden) instead: `array( 'status' => 403 )`.

**Minor** `misc/block-author-enumeration.php:24` — `! is_user_logged_in()` leaves logged-in subscribers able to enumerate authors. If the goal is to hide usernames from all non-editors, use `! current_user_can( 'edit_posts' )`. If logged-in enumeration is intentionally out of scope, the docblock should say so.

---

**Looks Good**

`misc/disable-user-sitemap.php` — Minimal and correct. Returning `false` from `wp_sitemaps_add_provider` is exactly the right pattern; no reinventing needed.

The REST API blocker's short-circuit `if ( null !== $result ) { return $result; }` at the top correctly respects prior middleware, which is easy to get wrong.

The coming-soon redirect handles all three loop-prevention cases cleanly (page not published → bail, already on page → bail, `wp_safe_redirect` + `exit`).

---

**Questions**

1. `rate-limit-password-resets.php` — Is `lostpassword_post` the intended hook? It fires before the email is sent, which is correct for blocking attempts. Just confirming that's intentional versus `retrieve_password` which fires after validation.

andrewlimaza:
@kimcoleman merging this in, looks good and is something that each site will need to tweak. You may also utilize the `pmpro_get_ip()` function which handles CloudFlare and other proxies automagically.

Sharing this for anyone that may stumble upon this PR and I've added a comment to using that function.

PR #464, opened 2026-05-27
Add customize-wordpress-password-reset-email.php

dwanjuki:
Migrating snippet to library: https://gist.github.com/kimwhite/30fb043d1b627baefa9067e88044a86b
https://trello.com/c/R5wtN8Qq/

- Added snippet metadata (title, layout, collection, category)
- Split docblock: file-level header separate from function docblock
- rawurlencode() the password reset key for URL safety
- Soften reset message copy to 'Someone has requested...'

### DIFF
--- a/email/customize-wordpress-password-reset-email.php
+++ b/email/customize-wordpress-password-reset-email.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Customize the WordPress password reset email.
+ *
+ * This is core WordPress functionality — not specific to Paid Memberships Pro.
+ *
+ * You can add this recipe to your site by creating a custom plugin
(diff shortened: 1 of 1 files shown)


kimcoleman:
This is similar to https://github.com/strangerstudios/pmpro-snippets-library/pull/470 - when we review this, we should choose one and close the other.

Note that I have a screenshot based on 470 in place on Trello for this companion blog post, so if that one is OK or just needs a small fixes, that would be nice to use :)

andrewlimaza:
Closing in favor of https://github.com/strangerstudios/pmpro-snippets-library/pull/470 (due to assets already being available). 🙏

PR #467, opened 2026-05-28
Create limit-editing-of-specific-posts-to-specific-users.php

dwanjuki:
Adding a snippet to protect post IDs from being edited: https://trello.com/c/8RG8gOoP/

### DIFF
--- a/misc/limit-editing-of-specific-posts-to-specific-users.php
+++ b/misc/limit-editing-of-specific-posts-to-specific-users.php
@@ -1,0 +1,57 @@
+<?php
+/**
+ * Only allow specific users to edit certain posts and pages.
+ *
+ * Map post IDs to an array of user IDs allowed to edit/delete/publish it.
+ * Any user not in the list will be blocked from those actions on that post.
+ * Being listed only exempts a user from this block.
+ * It does not grant users any capabilities they don't already have.
+ *
+ * title: Restrict Editing of Specific Posts to a Specific List of Users
+ * layout: snippet
+ * collection: misc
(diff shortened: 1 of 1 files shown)


flintfromthebasement (approved):
**PR: strangerstudios/pmpro-snippets-library#467 — Create limit-editing-of-specific-posts-to-specific-users.php**
*dwanjuki → dev | 1 file, +57 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/467>

---

**Summary**
New snippet that restricts edit/delete/publish capabilities on specific post IDs to a configured allowlist of user IDs. Clean, focused, correct. Ready to merge.

---

**Issues**

- **Minor** `limit-editing-of-specific-posts-to-specific-users.php:56` — `$caps[0]` is accessed without checking that `$caps` is non-empty. The `user_has_cap` filter receives `$caps` as the second argument (the primitive capabilities WP mapped from the meta cap), and it's almost always populated at this point — but a defensive `isset( $caps[0] )` before the assignment would be consistent with the early-exit pattern used for `$args` above.

---

**Looks Good**
- The strict `in_array(..., true)` comparisons on both `$cap` and `$user_id` prevent type-juggling surprises.
- Bailing early on non-allowlisted posts means this runs at minimal cost on every `user_has_cap` call.
- The comment that being in the allowlist *doesn't grant* capabilities the user doesn't already have is exactly the right expectation-setter — that distinction trips people up.
- `link: TBD` in the header is fine for pre-publish; just needs updating before the companion article goes live.

PR #469, opened 2026-05-29
Add Change Membership Level on Approval snippet

dwanjuki:
Migrating snippet to library: https://gist.github.com/MaryOJob/8dc3d7b20df14345570388b6b3ac4db5

Passing custom level instead of new level ID

Updated to prevent subscription cancellation on level change

### DIFF
--- a/add-ons/pmpro-approvals/change-level-on-approval.php
+++ b/add-ons/pmpro-approvals/change-level-on-approval.php
@@ -19,7 +19,7 @@ function my_pmpro_change_level_if_approved( $user_id, $level_id ) {
 	$new_level_id      = 2; // Level to change to after approval.
 
 	// Only run if the approved level is the targeted Approval level.
-	if ( $level_id === $approval_level_id ) {
+	if ( (int) $level_id === $approval_level_id ) {
 		$current_level = pmpro_getSpecificMembershipLevelForUser( $user_id, $approval_level_id );
 
 		if ( ! empty( $current_level ) ) {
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#469 — Add Change Membership Level on Approval snippet**
*dwanjuki → dev | 1 file, +63 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/469>

**Summary**

Adds a snippet that hooks `pmpro_approvals_after_approve_member` to transparently swap a user from an approval-gated level to a destination level, carrying over billing details and re-pointing any active subscription rather than cancelling it. The general approach is correct. One data integrity gap needs fixing before merge.

**Issues**

**Major** `change-level-on-approval.php:48` — `pmpro_changeMembershipLevel()` return value is never checked. If the level change fails (DB error, user already lost the level between the guard check and the call), execution falls through to the subscription re-assignment block. The subscription gets its `membership_level_id` updated to `$new_level_id` even though the user was never actually moved there — an orphaned subscription pointing to a level the user doesn't hold.

```php
// Current — no guard
pmpro_changeMembershipLevel( $custom_level, $user_id );

// Fix
if ( ! pmpro_changeMembershipLevel( $custom_level, $user_id ) ) {
    return;
}
```

**Minor** `change-level-on-approval.php:22` — Strict `===` comparison between `$level_id` (from the action hook) and `$approval_level_id = 1` (int literal). All current admin callers pass `intval()` so this won't fail in practice, but a loose `==` or explicit `(int)$level_id === $approval_level_id` would be safer if someone calls `approveMember()` programmatically with a string ID.

**Minor** `change-level-on-approval.php:61` — `$previous_subscription->save()` return value is silently discarded. If `save()` returns `false` (requires `gateway`, `gateway_environment`, and `subscription_transaction_id` to be set — they will be for any real subscription, but still), the failure goes unlogged. Add `error_log()` or a `do_action` to expose the failure.

**Looks Good**

- The `add_filter`/`remove_filter` wrapper around `pmpro_changeMembershipLevel` for `pmpro_cancel_previous_subscriptions` is the established PMPro pattern — correct and clean.
- Carrying over `billing_amount`, `cycle_number`, `cycle_period`, `billing_limit`, `trial_amount`, `trial_limit`, and the discount code from the approval level to the new level is the right call. A plain `pmpro_changeMembershipLevel( $new_level_id, $user_id )` would have reset billing details.
- Updating `membership_level_id` on the existing subscription rather than cancelling and creating a new one is the correct approach — avoids gateway-side cancellation noise.
- `startdate`/`enddate` conversion from Unix timestamp (what `pmpro_getSpecificMembershipLevelForUser` returns) to MySQL datetime string (what `pmpro_changeMembershipLevel` expects) is handled correctly.

PR #470, opened 2026-05-29
Customize the default WordPress password reset email 

kimwhite:
Adds a new snippet recipe that demonstrates how to customize the default WordPress password reset email using the retrieve_password_message filter.

This recipe is not specific to Paid Memberships Pro. It serves as an example for users who want to modify the content of the password reset email sent by WordPress, including the greeting, instructions, and messaging around the password reset link. The snippet provides a starting point that users can adapt to match their site's branding and communication style.

### DIFF
--- a/email/customize-wordpress-password-reset-email.php
+++ b/email/customize-wordpress-password-reset-email.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * Customize the WordPress password reset email sent to users.
+ *
+ * title: Customize the WordPress Password Reset Email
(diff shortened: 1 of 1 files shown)


kimcoleman:
This is similar to https://github.com/strangerstudios/pmpro-snippets-library/pull/464 - when we review this, we should choose one and close the other.

Note that I have a screenshot based on this one in place on Trello for this companion blog post, so if this one is OK or just needs a small fixes, it would be nice to use it :)

PR #473, opened 2026-06-04
Show the group leader's expiration or renewal date to child group members.

kimwhite:
Show the group leader's expiration or renewal date to child group members.

### DIFF
--- a/add-ons/pmpro-group-accounts/show-group-leader-expiration-renewal-date.php
+++ b/add-ons/pmpro-group-accounts/show-group-leader-expiration-renewal-date.php
@@ -58,14 +58,9 @@ function my_pmprogroupacct_membership_expiration_text( $text, $level, $user ) {
 	$group_leader_id = $group->group_parent_user_id;
 
 	// Check for a fixed expiration date on the group leader's membership.
-	$expiration_text = pmpro_get_membership_expiration_text(
-		$group->group_parent_level_id,
-		$group_leader_id,
-		''
-	);
-
-	if ( ! empty( $expiration_text ) && strpos( $expiration_text, 'No Expiration' ) === false ) {
-		return $expiration_text;
+	$leader_level = pmpro_getSpecificMembershipLevelForUser( $group_leader_id, $group->group_parent_level_id );
+	if ( ! empty( $leader_level->enddate ) ) {
+		return date_i18n( get_option( 'date_format' ), $leader_level->enddate );
 	}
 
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#473 — Show Leader date on Child account**
*kimwhite → dev | 1 file, +101 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/473>

**Summary**
New snippet that replaces the child member's expiration text with the group leader's fixed enddate or next subscription renewal date. Logic is sound and handles the three expected states (fixed enddate, active subscription, neither). Two issues worth fixing before merge: a recursive filter invocation that causes extra DB queries on every call site, and a dead code guard.

**Issues**

- **Major** `show-group-leader-expiration-renewal-date.php:62-67` — `pmpro_get_membership_expiration_text()` applies the `pmpro_membership_expiration_text` filter at its end (see `functions.php:704`). Calling it from *within* that filter means this function re-fires for the group leader on every invocation. It terminates in the normal case because the leader isn't also a child, but it silently executes `get_group_members()` for the leader on every call — including non-child members and every row in the Members List admin table. For a group with 50 members, that's 50 extra `get_group_members()` queries per page load. If the leader is ever a child of another group, the recursion goes three levels deep. Fix: inline the leader's enddate check directly instead of calling `pmpro_get_membership_expiration_text()`:

  ```php
  $leader_level = pmpro_getSpecificMembershipLevelForUser( $group_leader_id, $group->group_parent_level_id );
  if ( ! empty( $leader_level->enddate ) ) {
      return date_i18n( get_option( 'date_format' ), $leader_level->enddate );
  }
  ```

- **Minor** `show-group-leader-expiration-renewal-date.php:65` — `strpos( $expiration_text, 'No Expiration' ) === false` is dead code. `pmpro_get_membership_expiration_text()` with `$default = ''` returns `''` when there's no enddate — never the string "No Expiration". The `empty( $expiration_text )` check already handles the no-enddate case correctly. Remove the `strpos` guard.

- **Minor** `show-group-leader-expiration-renewal-date.php:101` — missing newline at end of file.

**Looks Good**
The class existence checks before doing anything (`class_exists( 'PMProGroupAcct_Group_Member' )`) are the right pattern for a snippet that depends on an add-on. The direct DB query on `pmpro_subscriptions` uses `wpdb->prepare()` correctly with `%d`, and `next_payment_date` is a real indexed column in that table.

PR #475, opened 2026-06-12
Allow Pending Members to see non-member content

kimwhite:
This is used to modify the behavior described here: https://github.com/strangerstudios/pmpro-approvals/issues/192

This recipe allow the pending member to view non-member content block or shortcode content.

The final version keeps the original behavior but adds a few safeguards and makes the membership-level check more flexible.

**Changes from the Original**

**Added an admin check earlier in the function**

The original snippet skipped admin requests after several other checks.
The final version exits immediately on admin requests, avoiding unnecessary processing.

**Made the PMPro page check safer**

The original code used `is_page( $pmpro_pages )`.
The final version only runs the page check after WordPress has finished loading the current page `(did_action( 'wp' )) `and `uses array_values( $pmpro_pages )` to ensure only page IDs are checked.
This prevents issues if the filter runs before the queried page is available.

**Added a check to confirm the Approvals Add On is active**

The final version verifies that the `PMPro_Approvals` class exists before attempting to use it.
This prevents fatal errors if the Add On is deactivated.

**Improved detection of "non-member content" checks**

The original snippet only handled cases where `$levels` was an array containing a single value of 0.
The final version also supports checks where `$levels `is passed directly as `0` or `'0'`.
This makes the snippet compatible with more ways developers may call `pmpro_hasMembershipLevel().`

**Simplified the approval check**

The original version created an instance of the Approvals class before checking the member status.
The final version calls `PMPro_Approvals::isPending() `directly, which is cleaner and easier to read.
Result

The functionality remains the same: users with a single pending approval level will see content intended for non-members.

The final version is more defensive, handles additional edge cases, and is less likely to cause issues when used alongside custom code or future PMPro updates.

### DIFF
--- a/add-ons/pmpro-approvals/allow-pending-members-to-view-non-member-content.php
+++ b/add-ons/pmpro-approvals/allow-pending-members-to-view-non-member-content.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * Allow pending approval members to see content shown to non-members.
+ *
+ * title: Allow Pending Approval Members to View Non-Member Content
+ * layout: snippet-example
+ * collection: content-controls
+ * category: approvals
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+/**
+ * This adjusts checks for level ID 0, which PMPro uses when checking
+ * whether a user should see non-member content. If the logged-in user
+ * only has one membership level and that level is pending approval,
+ * this returns true so non-member content is shown.
+ */
+function my_pmproapprovals_show_non_member_content_to_pending_users( $has_level, $user_id, $levels ) {
+	global $pmpro_pages;
+
+	// Do not adjust checks in the WordPress admin.
+	if ( is_admin() ) {
+		return $has_level;
+	}
+
+	// Only adjust checks that are currently returning false.
+	if ( $has_level ) {
+		return $has_level;
+	}
+
+	// Let PMPro handle access checks on PMPro-generated pages.
+	if ( did_action( 'wp' ) && ! empty( $pmpro_pages ) && is_page( array_values( $pmpro_pages ) ) ) {
+		return $has_level;
+	}
+
+	// A user ID is required to check approval status.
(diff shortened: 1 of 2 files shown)


flintfromthebasement (approved):
**PR: strangerstudios/pmpro-snippets-library#475 — Allow Pending Members to see non-member content**
*kimwhite → dev | 1 file, +75 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/475>

**Summary**
Adds a snippet that hooks `pmpro_has_membership_level` at priority 15 to return `true` for non-member content checks (`$levels === 0`) when the requesting user has exactly one pending-approval level. Logic, guard ordering, and filter priority are all correct — runs after the Approvals plugin's own priority 10 filter, which short-circuits on `! $haslevel` and hands control cleanly to this snippet. Ready to merge with one small defensive fix and a nit.

**Issues**

- **Minor** `allow-pending-members-to-view-non-member-content.php:62` — `pmpro_getMembershipLevelsForUser()` can return `false` if `$user_id` is somehow empty despite the earlier guard (e.g., a third-party plugin calls the filter with a falsy but non-empty `$user_id` that slips past `empty()`). Calling `count(false)` is a `TypeError` in PHP 8. Change:
  ```php
  if ( count( $user_levels ) === 1 && PMPro_Approvals::isPending( ... ) ) {
  ```
  to:
  ```php
  if ( is_array( $user_levels ) && count( $user_levels ) === 1 && PMPro_Approvals::isPending( ... ) ) {
  ```

- **Minor** `allow-pending-members-to-view-non-member-content.php:75` — Missing newline at end of file (the diff shows `\ No newline at end of file`).

**Looks Good**

- Guard ordering is solid and deliberate: admin bail-out → already-true short-circuit → PMPro page passthrough → empty user → class exists → non-member check → single pending level. Each gate is justified.
- Using `array_values( $pmpro_pages )` before `is_page()` is cleaner than passing the keyed associative array directly (even though WordPress handles both).
- `PMPro_Approvals::isPending()` is the right call — it handles the "level existed before approval was required" edge case internally via `hasMembershipLevelSansApproval()`, so the snippet doesn't need to replicate that logic.
- Priority 15 after the Approvals plugin's priority 10 is the correct placement. The Approvals filter returns early on `! $haslevel` for level-0 checks, leaving this snippet to make the final call cleanly.

**Questions**

1. Should this also handle users with multiple levels where all of them are pending? The current `count === 1` requirement means a user who signed up for two levels simultaneously (both pending) won't see non-member content. Intentional?

PR #479, opened 2026-06-19
Add and modernize getfile protect non-WordPress files snippet

kimcoleman:
## Summary

Adds the **getfile-protect-non-wordpress-files** snippet to the library and brings it up to current standards. This recipe serves files from a protected directory only to members, redirecting non-members to the login page.

## Changes

**Correctness / security**
- Renamed the deprecated `pmpro_getfile_extension_blacklist` filter to `pmpro_getfile_extension_blocklist` (the old name is deprecated in core's `includes/deprecated.php`).
- Removed the duplicate `require_once` for the mimetype class.
- Pass explicit level IDs to `pmpro_hasMembershipLevel( array( 1, 2 ) )` instead of allowing **any** active level (the old comment said "level 2 or 3" but the code passed no args).
- Replaced the `is_admin()` bypass with a real `current_user_can( 'manage_options' )` capability check (`is_admin()` only checks admin-page context, not the user's role).
- Switched `$_REQUEST` to `$_GET` and sanitize the input with `wp_unslash()` + `sanitize_text_field()`.
- Build redirects with `home_url()` instead of manually assembling scheme + `HTTP_HOST`.
- Removed the unused `$wpdb` global and the unused `$pathParts` variable.

**Standards**
- Converted the header to the library frontmatter format (`title` / `layout` / `collection` / `category` + companion-article boilerplate), keeping the `.htaccess` and setup instructions.
- Setup notes now reference the `PROTECTED_DIR` constant and `pmpro_hasMembershipLevel()` call by name instead of brittle line numbers.
- Applied WordPress coding standards: single quotes, spacing inside parens, Yoda conditions, strict comparisons (`===`, `in_array( ..., true )`), braces on the same line, and an early return to reduce nesting.

The blocklist fallback redirect (to the direct file URL with `?noloop=1`) is intentional and matches core behavior — left as-is.

`php -l` passes.

### DIFF
--- a/restricting-content/getfile-protect-non-wordpress-files.php
+++ b/restricting-content/getfile-protect-non-wordpress-files.php
@@ -1,0 +1,123 @@
+<?php
+/**
+ * Protect non-WordPress files in a directory and only serve them to members.
+ * Loads a file from a protected directory (e.g. /protected-directory/) and
+ * redirects non-members to the login page before the file is served.
+ *
+ * title: Protect Non-WordPress Files in a Directory for Members Only
+ * layout: snippet
+ * collection: restricting-content
+ * category: content, restriction, member-access
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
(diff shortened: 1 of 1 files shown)


kimcoleman:
Changed this to a draft as we may be totally replacing this with #480

andrewlimaza:
Closing in favor of #480 as discussed.

PR #480, opened 2026-06-19
Add getfile snippet: serve a protected folder inline to members (finfo/inline)

kimcoleman:
## Summary

Adds the modernized **getfile** recipe for serving a folder of non-WordPress files (e.g. an AI-generated artifact — `index.html` + JSON/CSS/JS/images) to members only. This is the version that pairs with the in-place refresh of [Locking Down Non-WordPress Files and Folders with PMPro](https://www.paidmembershipspro.com/locking-non-wordpress-files-folders-paid-memberships-pro/).

Files are served **inline** so the artifact renders in the browser (and inside an iframe) instead of downloading.

## How it works

- Pairs with a rewrite rule that routes `/protected-tools/` requests to `index.php?pmpro_getfile={path}` (Apache + NGINX rules documented in the snippet header).
- Real files live in `wp-content/uploads/protected-tools/` (or anywhere readable, including outside the web root).
- **Traversal protection:** strips `../`-style sequences, then resolves with `realpath()` and verifies the result is still inside the protected folder via a `DIRECTORY_SEPARATOR`-anchored prefix check.
- **MIME detection:** maps common web types (`html`, `css`, `js`, `json`, `svg`) by extension first — because `finfo` reports `.css`/`.js` as `text/plain`, which browsers won't apply — then falls back to `finfo_open( FILEINFO_MIME_TYPE )` for everything else. Requires the PHP `fileinfo` extension.
- **Membership check:** `pmpro_hasMembershipLevel()` (allows any active member); inline docs show how to restrict to specific level IDs.
- Library frontmatter + WordPress coding standards (tabs).

## Note for reviewers

This is a **different implementation** from open PR #479 (which cleaned up the legacy web-root + `pmpro_mimetype` + extension-blocklist version). Both currently target the same file path, `restricting-content/getfile-protect-non-wordpress-files.php`. We can decide which approach the library should ship and close/rework the other — see PR #479 for the comparison.

`php -l` passes.

### DIFF
--- a/restricting-content/getfile-protect-non-wordpress-files.php
+++ b/restricting-content/getfile-protect-non-wordpress-files.php
@@ -37,8 +37,20 @@ function my_pmpro_getfile() {
 		return;
 	}
 
+	// Membership check first, before any filesystem work. Checking here (rather
+	// than after resolving the path) means non-members always get a 403 whether or
+	// not the requested file exists, so the response can't be used to probe for it.
+	// With no arguments this allows any active member. To limit to specific levels,
+	// pass their IDs, e.g.: if ( ! pmpro_hasMembershipLevel( array( 3, 7, 10 ) ) ).
+	if ( ! pmpro_hasMembershipLevel() ) {
+		wp_die( esc_html__( 'This file is available to members only.', 'your-text-domain' ), '', array( 'response' => 403 ) );
+	}
+
 	// Where the real files live. This can be anywhere your server can read,
 	// including outside the web root. Here we use a folder in wp-content/uploads.
+	// Only let trusted admins write to this folder: a malicious file placed here
+	// (e.g. an SVG containing <script>) would be served to members and could run
+	// JavaScript in their browser.
 	$base_dir  = trailingslashit( wp_upload_dir()['basedir'] ) . 'protected-tools';
 	$real_base = realpath( $base_dir );
 
@@ -53,13 +65,6 @@ function my_pmpro_getfile() {
 		|| strpos( $file_path, $real_base . DIRECTORY_SEPARATOR ) !== 0
 		|| ! is_file( $file_path ) ) {
 		wp_die( esc_html__( 'File not found.', 'your-text-domain' ), '', array( 'response' => 404 ) );
-	}
-
-	// Membership check. With no arguments this allows any active member.
-	// To limit to specific levels, pass their IDs, e.g.:
-	//   if ( ! pmpro_hasMembershipLevel( array( 3, 7, 10 ) ) ) { ... }
-	if ( ! pmpro_hasMembershipLevel() ) {
-		wp_die( esc_html__( 'This file is available to members only.', 'your-text-domain' ), '', array( 'response' => 403 ) );
 	}
 
 	// Figure out the MIME type so the browser handles the file correctly.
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#480 — Add getfile snippet: serve a protected folder inline to members (finfo/inline)**
*kimcoleman → dev | 1 file, +96 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/480>

---

**Summary**

Clean, well-documented new snippet. The path traversal protection is solid — `realpath()` + the `DIRECTORY_SEPARATOR`-anchored prefix check is the real gate, and the comment explains why the trailing separator matters. Two issues to fix before shipping, one of them a meaningful security concern.

---

**Issues**

**Major** `getfile-protect-non-wordpress-files.php:55–60` — The membership check runs *after* file path resolution and existence validation. Unauthenticated users can probe for file existence: a request for a file that doesn't exist returns 404, a request for one that does returns 403. Move the `pmpro_hasMembershipLevel()` check above the path traversal block so non-members always get 403 regardless of whether the file exists.

```php
// Do this first, before any path work:
if ( ! pmpro_hasMembershipLevel() ) {
    wp_die( esc_html__( 'This file is available to members only.', 'your-text-domain' ), '', array( 'response' => 403 ) );
}
```

**Minor** `getfile-protect-non-wordpress-files.php:73–78` — `finfo_open()` can return `false` if the `fileinfo` extension is unavailable. Passing `false` as the first argument to `finfo_file()` will PHP-fatal, leaving users with a blank page. Add a fallback:

```php
$finfo = finfo_open( FILEINFO_MIME_TYPE );
if ( false === $finfo ) {
    $content_type = 'application/octet-stream';
} else {
    $content_type = finfo_file( $finfo, $file_path );
    finfo_close( $finfo );
}
```

**Minor** `getfile-protect-non-wordpress-files.php:65` — SVG is in the known-types map and gets served as `image/svg+xml` inline. SVG files can contain `<script>` tags and execute JavaScript in the browser when served with that MIME type. Since this is member-only, the blast radius is limited — a member would have to upload a malicious SVG to the protected folder — but if the protected folder is writable by anyone other than the site admin, this is an XSS vector. Worth noting in the snippet's inline docs.

---

**Looks Good**

The `realpath()` + `DIRECTORY_SEPARATOR`-anchored prefix check (`strpos($file_path, $real_base . DIRECTORY_SEPARATOR) !== 0`) is the correct way to prevent traversal — the comment explaining *why* the trailing separator matters is exactly the right kind of non-obvious note. The extension-first MIME map with the `finfo` fallback rationale is also clearly explained and correctly addresses the `text/plain` problem for `.css`/`.js`.

---

**Questions**

1. This PR and #479 both write to `restricting-content/getfile-protect-non-wordpress-files.php`. Which one is the canonical path going forward? The PR description flags this but doesn't resolve it — closing one or renaming one of the two files before merge would avoid a last-writer-wins conflict.

kimcoleman:
Feedback from Flint's review is processed:

1. Major (existence oracle) — The pmpro_hasMembershipLevel() block now runs immediately after the empty() early
return, before any path/realpath/existence work. Non-members get a uniform 403 regardless of whether the file
exists, and we skip filesystem work for rejected requests. Kept the "restrict to specific levels" inline doc
with it.

2. Minor (finfo_open() false) — Wrapped it: if finfo_open() returns false (no fileinfo extension), it falls
back to application/octet-stream instead of fataling on finfo_file( false, ... ). As noted, this only affects
non-mapped types (images/PDFs/fonts) since html/css/js/json/svg come from the extension map.

3. Minor (SVG XSS) — Added the caveat to the $base_dir comment: only trusted admins should be able to write to
the protected folder, since a malicious SVG with <script> would be served inline to members and execute JS.
Left SVG in the map (removing it wouldn't help — finfo would still serve it as image/svg+xml).

PR #483, opened 2026-06-23
Add German Widerruf/Kündigung cancellation buttons snippet

kimcoleman:
## What

Adds a new snippet — `membership-levels/add-german-widerruf-kuendigung-buttons.php` — that adapts PMPro's native cancel flow to the German Widerrufs-/Kündigungsbutton rules (§ 356a, § 312k). For a blog post.

It reuses the existing per-membership account action link, the cancel confirmation page, and the cancellation emails — no parallel UI — and layers on:

1. **Account page:** a "Vertrag widerrufen" entry link per membership and relabels the existing cancel link to "Verträge hier kündigen" (`pmpro_member_action_links`).
2. **Withdrawal flag:** a hidden field carried through the confirmation form so later steps know this was a withdrawal (`pmpro_cancel_before_submit`).
3. **Confirmation button labels:** "Widerruf bestätigen" / "jetzt kündigen", filtered on `ngettext` since core builds the button with `_n()`.
4. **Immediate cancellation** for a withdrawal instead of running to the next payment date (`pmpro_cancel_on_next_payment_date`); a normal Kündigung keeps PMPro's default behavior.
5. **`!!cancellation_datetime!!`** email variable to record receipt date/time (`pmpro_email_data`).
6. **Optional confirmation email address** for the withdrawal acknowledgment, per § 356a (2) (`pmpro_cancel_before_submit` + `pmpro_email_recipient`). Member-facing templates only; gated to withdrawals.

## Placement

`membership-levels` / category `cancellation` — the defining behavior is the cancellation flow (immediate vs. run-to-term), matching the existing `cancel-*-immediately.php` snippets.

## Verification

All hooks, filter signatures, the `_n()` button strings, and the `pmpro_url` query format were checked against PMPro core. Steps 4 and 6 change cancellation timing and email delivery — test on staging and confirm wording with a legal-text provider before production. Labels are German; swap for locale. `link:` header is `TBD` pending the blog post URL.

### DIFF
--- a/membership-levels/add-german-widerruf-kuendigung-buttons.php
+++ b/membership-levels/add-german-widerruf-kuendigung-buttons.php
@@ -25,11 +25,11 @@
  *    the existing cancel link to "Verträge hier kündigen".
  */
 function my_pmpro_widerruf_action_links( $links, $level_id ) {
-	$widerruf_url = pmpro_url( 'cancel', 'levelstocancel=' . $level_id . '&widerruf=1' );
+	$widerruf_url = pmpro_url( 'cancel', 'levelstocancel=' . intval( $level_id ) . '&widerruf=1' );
 	$links['widerruf'] = '<a id="pmpro_actionlink-widerruf" href="' . esc_url( $widerruf_url ) . '">' . esc_html( 'Vertrag widerrufen' ) . '</a>';
 
 	if ( isset( $links['cancel'] ) ) {
-		$cancel_url      = pmpro_url( 'cancel', 'levelstocancel=' . $level_id );
+		$cancel_url      = pmpro_url( 'cancel', 'levelstocancel=' . intval( $level_id ) );
 		$links['cancel'] = '<a id="pmpro_actionlink-cancel" href="' . esc_url( $cancel_url ) . '">' . esc_html( 'Verträge hier kündigen' ) . '</a>';
 	}
 
@@ -58,6 +58,9 @@ function my_pmpro_confirm_button_label( $translation, $single, $plural, $number,
 	}
 
 	// Singular forms of the confirm button for specific levels and for all levels.
+	// These must match the _n() singular strings in paid-memberships-pro/pages/cancel.php
+	// (the "Yes, cancel this membership" and "Yes, cancel my membership" calls). If PMPro
+	// renames them, this filter silently stops matching — re-verify against core on upgrade.
 	$confirm_singulars = array(
 		'Yes, cancel this membership',
 		'Yes, cancel my membership',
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#483 — Add German Widerruf/Kündigung cancellation buttons snippet**
*kimcoleman → dev | 1 file, +144 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/483>

---

**Summary**
Solid snippet with a clear legal use case. All hook signatures verified against PMPro core (`pmpro_cancel_on_next_payment_date`, `pmpro_email_recipient`, `pmpro_cancel_before_submit`, `pmpro_member_action_links` — all correct). Output is consistently escaped. The cancel timing change (`pmpro_cancel_on_next_payment_date`) runs inside PMPro's nonce-verified flow, so no CSRF exposure. Two issues to address before merge, both fixable without redesign.

---

**Issues**

**Major** `add-german-widerruf-kuendigung-buttons.php:63` — The `ngettext` filter is registered unconditionally on every page load. `ngettext` fires on every `_n()` call across the entire site — not just the cancel page. Every PMPro-domain plural string on every page (account, checkout, admin) executes this function. Scope it to the cancel page:

```php
add_action( 'wp', function() {
    if ( is_page( pmpro_getOption( 'cancel_page_id' ) ) ) {
        add_filter( 'ngettext', 'my_pmpro_confirm_button_label', 10, 5 );
    }
} );
```

**Major** `add-german-widerruf-kuendigung-buttons.php:68-72` — The `ngettext` filter matches on hardcoded English strings from PMPro core (`'Yes, cancel this membership'`, `'Yes, cancel my membership'`). If PMPro renames these in a future version, the filter silently stops matching and the button stays in English — no error, no warning. Add a code comment:

```php
// These strings must match the singular forms in paid-memberships-pro/pages/cancel.php _n() calls.
// Verify against PMPro core when upgrading.
```

**Minor** `add-german-widerruf-kuendigung-buttons.php:27` — `'levelstocancel=' . $level_id` passes `$level_id` raw into the URL query string. PMPro always passes an integer here, but `intval( $level_id )` makes the safety explicit. Low actual risk.

**Minor** `add-german-widerruf-kuendigung-buttons.php:129-131` — `is_email( $_REQUEST['widerruf_confirmation_email'] )` validates the raw slashed value; `sanitize_email( wp_unslash( ... ) )` correctly unslashes before sanitizing. Backslashes don't affect `is_email()` in practice, but applying `wp_unslash()` before both checks is consistent.

---

**Looks Good**
- Hook 6b correctly validates with `is_email()` + sanitizes with `sanitize_email( wp_unslash() )` before redirecting the email recipient — that's the right sequence.
- Scoping the recipient override to member-only templates (`cancel`, `cancel_on_next_payment_date`) and never the `_admin` copies is exactly right.
- `wp_date()` with no timestamp in hook 5 correctly captures request time using site date/time format options.

---

**Questions**
1. The PR description says steps 4 and 6 need staging verification — is there a staging test report, or is that expected to come from the blog post author before publish?
2. Does this snippet handle the multi-level case cleanly? The account page `widerruf` link is per-membership (`levelstocancel=$level_id`), but if a member has multiple levels and navigates to the "all" cancel flow, `levelstocancel` is `all` — `empty($_REQUEST['levelstocancel'])` would be false, so the `ngettext` filter would fire. Does the `widerruf` flow make sense when cancelling all levels simultaneously?

kimcoleman:
Thanks for the review. Pushed cc21663 addressing 3 of the 4 items; declining one with reasoning below.

**Issue 2 (Major) — addressed.** Verified both confirm-button strings match PMPro core exactly (`pages/cancel.php` lines 56 and 64). Added a comment documenting that they must track the `_n()` singular forms in core, and to re-verify on upgrade.

**Issue 4 (Minor) — addressed.** Now unslash + `sanitize_email()` once into a local `$confirmation_email`, then run `is_email()` on the clean value, so both the presence check and validation operate on the unslashed value.

**Issue 3 (Minor) — addressed, both sites.** Wrapped the level id in `intval()`. Note `levelstocancel=` is built in two places (the widerruf link and the relabeled cancel link), so I applied it to both.

**Issue 1 (Major) — declining the rewrite.** The `ngettext` callback already early-returns on `empty( $_REQUEST['levelstocancel'] )` — a domain string compare plus an `empty()` check, so off the cancel page it's effectively free; the per-call cost doesn't justify a "Major." More importantly, the suggested `pmpro_getOption( 'cancel_page_id' )` relies on a function PMPro core documents as "Will be deprecated in a future release" (`includes/functions.php`), so it would trade a trivial guard for a deprecating dependency. The existing `levelstocancel` check is also the correct scoping signal, since the filter only ever needs to act when that request var is present.

PR #485, opened 2026-06-24
Add lead scoring CRM report snippet

dwanjuki:
Adds `misc/pmpro-lead-scoring-crm-report.php`, a snippet that turns a PMPro site into a lightweight lead-scoring CRM:

- **Tracks** high-intent page views for logged-in members (`template_redirect`).
- **Scores** each member from a profile "stage" field + weighted page views + phone-on-file + paying status, stored in `lead_score` user meta (recalculated on `pmpro_after_checkout` and `profile_update`).
- **Reports** via a new "Lead Scoring" page under *Memberships → Reports*, with a ranked member table and an editable, nonce-protected sales-pipeline status.

### Open items (non-blocking)
- `link:` header is still `TBD` — needs the companion article URL.
- `category:` is `reports, crm` — confirm whether the library supports multiple categories or pick one.
- Setup steps (create the two user fields, match the Select option values, replace the dog-training placeholder content) to be covered in docs.

### DIFF
--- a/misc/pmpro-lead-scoring-crm-report.php
+++ b/misc/pmpro-lead-scoring-crm-report.php
@@ -332,20 +332,17 @@ function pmpro_report_pmprocrm_page() {
 							</select>
 						</form>
 					</td>
+					<?php
+					$member_url = add_query_arg(
+						array(
+							'page'    => 'pmpro-member',
+							'user_id' => $user_id,
+						),
+						admin_url( 'admin.php' )
+					);
+					?>
 					<td>
-						<strong><a href="
-						<?php
-						echo esc_url(
-							add_query_arg(
-								array(
-									'page'    => 'pmpro-member',
-									'user_id' => $user_id,
-								),
-								admin_url( 'admin.php' )
-							)
-						);
-						?>
-											"><?php echo esc_html( $user->display_name ); ?></a></strong><br />
+						<strong><a href="<?php echo esc_url( $member_url ); ?>"><?php echo esc_html( $user->display_name ); ?></a></strong><br />
 						<?php echo esc_html( $user->user_email ); ?><br />
 						<?php echo ! empty( $level ) ? 'Level: ' . esc_html( $level->name ) : 'No level'; ?>
 						<?php
--- a/misc/pmpro-lead-scoring-crm-report.php
+++ b/misc/pmpro-lead-scoring-crm-report.php
@@ -14,9 +14,11 @@
  * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
  */
 
-/* ===========================================================================
+/*
+ * ===========================================================================
  * CONFIG — the only things you'd change for your own site.
- * ======================================================================== */
+ * ===========================================================================
+ */
 
 // The meta keys of the two fields you created in the GUI.
 define( 'PMPROCRM_PHONE_META', 'phone' );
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#485 — Add lead scoring CRM report snippet**
*dwanjuki → dev | 1 file, +336 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/485>

**Summary**
Solid, well-structured snippet — nonce/cap checks on the mutating action are correct, and the `pmpro_reports`/`pmpro_report_<key>_page`/`_widget` wiring matches core's actual convention (verified against `adminpages/reports.php`). One real perf/data-integrity issue worth fixing before merge; the rest are minor.

**Issues**

- **Major** `misc/pmpro-lead-scoring-crm-report.php:242` — `pmprocrm_calculate_lead_score( $user_id )` is called inside the render loop for every row (up to 200 rows per page load), and that function ends with `update_user_meta( $user_id, 'lead_score', $score )`. Every time an admin *views* the report, it writes to `usermeta` for every visible member — a write-on-read side effect that also makes page loads slower than necessary. The query already fetches `um_score.meta_value` — use that stored value for display and only recalculate on the actual trigger hooks (`pmpro_after_checkout`, `profile_update`), or a dedicated recompute action.

- **Minor** `misc/pmpro-lead-scoring-crm-report.php:96` — `update_user_meta( $user_id, 'lead_score', $score )` uses an unprefixed generic meta key (`lead_score`) while everything else in the snippet uses the `pmprocrm_` prefix (`pmprocrm_status`, `pmprocrm_page_views`). A bare `lead_score` key is likely to collide with another plugin or snippet on the same site. Rename to `pmprocrm_lead_score` for consistency and to reduce collision risk.

- **Minor** `misc/pmpro-lead-scoring-crm-report.php:210` — `wp_verify_nonce( $_GET['_wpnonce'], ... )` passes the raw superglobal value instead of `sanitize_text_field( wp_unslash( $_GET['_wpnonce'] ) )`, unlike the `status` field two lines below which is properly unslashed/sanitized. Low risk since `wp_verify_nonce` doesn't do anything dangerous with a raw string, but it's an inconsistency worth cleaning up.

- **Minor** `misc/pmpro-lead-scoring-crm-report.php:222` — The report query joins `mu.status = 'active'` only, so trialing/pending or cancelled-but-recent members never surface regardless of lead score. That may be intentional for a "current members" CRM view, but worth a one-line comment or a config toggle since a high-scoring lead who hasn't converted yet (still on a trial, or between charges) is exactly the audience a lead-scoring tool should surface.

- **Minor** `misc/pmpro-lead-scoring-crm-report.php:158` — `pmprocrm_get_main_level_for_user()` takes `reset( $levels )` (first level PMPro returns) for multi-level members. Fine as a "good enough for a demo" call per the docblock, but flag it so anyone customizing for a multi-level site knows this is a simplification, not a guarantee of "primary" level.

**Looks Good**
Nonce + `current_user_can( 'edit_user', $user_id )` ownership check on the status-update handler is correctly ordered (nonce first, then capability, then whitelist-check the status value against `pmprocrm_statuses()`) — no open write path there. Escaping throughout the table render (`esc_html`, `esc_attr`, `esc_url`) is consistent and correct.

**Questions**
1. Is the write-on-view recalculation in the loop intentional (e.g., to self-heal missing meta for pre-existing members), or an oversight? If intentional, worth a comment explaining why.
2. Given the `about` page is scored highest (8) as "the premium 1:1 offer," should the CONFIG block note that page weights are meant to reflect the sales funnel structure of the *specific* site, not a general default?

flintfromthebasement (approved):
**Follow-up review**

**PR: strangerstudios/pmpro-snippets-library#485 — Add lead scoring CRM report snippet**
*dwanjuki → dev | 1 file, +336 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/485>

**Summary**
All prior blocking and cleanup points are resolved; this clears my last review. Ready to merge.

**Prior issues — status**

- **Resolved** — `pmprocrm_calculate_lead_score()` no longer runs in the report render loop. The report now reads the stored `pmprocrm_lead_score` meta directly (`misc/pmpro-lead-scoring-crm-report.php:311`), and recalculation moved to the actual trigger points: `pmpro_after_checkout`, `profile_update`, and now also `pmprocrm_track_page_views()` on a genuine tracked-page view (`:124`, gated inside the `if ( empty( $current ) ) { return; }` branch — confirmed it only fires when `$current` is a real tracked slug, not on every page load). Write-on-read is gone.
- **Resolved** — `lead_score` renamed to `pmprocrm_lead_score` everywhere, including the SQL join (`:270`). Prefixed and consistent now.
- **Resolved** — nonce is sanitized before verification: `wp_verify_nonce( sanitize_text_field( wp_unslash( $_GET['_wpnonce'] ) ), ... )` (`:227`).
- **Not addressed** — the report query still filters to `mu.status = 'active'` only, so trialing/pending members never surface. Still Minor, still worth a comment or toggle if intentional — not blocking.
- **Not addressed** — `pmprocrm_get_main_level_for_user()` still takes `reset( $levels )` for multi-level members. Still Minor, already flagged as an acceptable simplification per the docblock.

**New since last review**
- Good catch (unprompted): the status-update `<select>` was renamed from `name="status"` to `name="new_status"` (`:335`), with a new hidden `status` field carrying the current tab filter (`:333`). In the prior version both the per-row status-change field and the report's tab-filter query var were named `status` — submitting a row's status form would have also clobbered the active tab filter on redirect. This fixes that latent bug as a side effect of the refactor.
- Cosmetic only: the `<a href="...">` block around `:349` now spans the PHP `esc_url()` call across multiple lines, so the rendered `href` attribute carries leading/trailing whitespace and tabs. Browsers trim this on load so it's not a functional bug — not worth blocking on, just flagging in case it's an artifact of an auto-formatter (phpcbf) run rather than intentional.

**Looks Good**
Clean resolution of the write-on-view issue — using the stored score for display while triggering recompute only from real scoring events is exactly the right fix. The `status`/`new_status` field disambiguation is a nice unprompted bug fix.

PR #487, opened 2026-06-26
Create show-all-member-markers-on-map-when-no-search-is-active.php

dwanjuki:
When the directory is viewed without a search query, the map normally only shows markers for the current paginated page of members. 

This snippet queries all active members with a saved map location and overrides the marker data before Google Maps initializes, so every member appears as a pin regardless of pagination.

### DIFF
--- a/add-ons/pmpro-member-directory/show-all-member-markers-on-map-when-no-search-is-active.php
+++ b/add-ons/pmpro-member-directory/show-all-member-markers-on-map-when-no-search-is-active.php
@@ -1,0 +1,115 @@
+<?php
+/**
+ * Show ALL member markers on the directory map when no search is active.
+ *
+ * When the directory is viewed without a search query, the map normally only shows
+ * markers for the current paginated page of members. This snippet queries all active
+ * members with a saved map location and overrides the marker data before Google Maps
+ * initializes, so every member appears as a pin regardless of pagination.
+ *
+ * title: Show All Member Markers on the Directory Map When No Search Is Active
+ * layout: snippet
+ * collection: pmpro-member-directory
+ * category: maps, markers
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+// Capture the exact shortcode attributes used when the map renders,
+// so we build marker info-windows with the same fields/links as the live map.
+function custom_pmpromd_capture_map_attributes( $map_id, $attributes ) {
+	global $custom_pmpromd_map_attributes;
+	$custom_pmpromd_map_attributes = $attributes;
+	return $map_id;
+}
+add_filter( 'pmpromd_map_id', 'custom_pmpromd_capture_map_attributes', 10, 2 );
+
+// Before footer scripts are printed, replace the paginated marker set with all members.
+function custom_pmpromd_all_markers_no_search() {
+	global $wpdb, $custom_pmpromd_map_attributes;
+
+	// Only act when the map script was actually enqueued on this page load.
+	if ( ! wp_script_is( 'pmpromd-google-maps-javascript', 'enqueued' ) ) {
+		return;
+	}
+
+	// Leave the default behaviour in place when the visitor has searched.
+	if ( ! empty( $_REQUEST['ps'] ) ) {
+		return;
+	}
+
+	// The plugin's marker-data generator must be available.
+	if ( ! function_exists( 'pmpromd_generate_marker_data' ) ) {
+		return;
+	}
+
+	$attributes = ! empty( $custom_pmpromd_map_attributes ) ? $custom_pmpromd_map_attributes : array();
+
+	// Optional: carry level restrictions from the directory shortcode into this query.
+	$level_sql = '';
+	if ( ! empty( $attributes['level'] ) ) {
+		$levels = is_array( $attributes['level'] ) ? implode( ',', $attributes['level'] ) : $attributes['level'];
+		$levels = preg_replace( '/[^0-9,]/', '', $levels );
+		if ( ! empty( $levels ) ) {
+			$level_sql = "AND mu.membership_id IN ($levels)";
+		}
+	}
+
+	// phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+	$sql = "SELECT
+			u.ID,
+			u.user_nicename,
+			umf.meta_value   AS first_name,
+			uml.meta_value   AS last_name,
+			ummap.meta_value AS maplocation
+		FROM {$wpdb->users} u
+		LEFT JOIN {$wpdb->pmpro_memberships_users} mu
+			ON u.ID = mu.user_id
+		LEFT JOIN {$wpdb->usermeta} umh
+			ON umh.meta_key = 'pmpromd_hide_directory' AND u.ID = umh.user_id
+		LEFT JOIN {$wpdb->usermeta} umf
+			ON umf.meta_key = 'first_name' AND u.ID = umf.user_id
+		LEFT JOIN {$wpdb->usermeta} uml
+			ON uml.meta_key = 'last_name' AND u.ID = uml.user_id
+		LEFT JOIN {$wpdb->usermeta} ummap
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#487 — Create show-all-member-markers-on-map-when-no-search-is-active.php**
*dwanjuki → dev | 1 file, +115 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/487>

---

**Summary**

New snippet for pmpro-member-directory that overrides the paginated marker set with all active members when no search is active. The core logic is sound — the filter hook, function existence guards, and JS injection point are all correct. Two issues need addressing before publish: a silent failure path in the JSON encoding step, and the absence of any scale warning for directories with many members. Plus a couple of minor cleanups.

---

**Issues**

- **Major** `show-all-member-markers-on-map-when-no-search-is-active.php:104` — `wp_json_encode` can return `false` on encoding failure (e.g. non-UTF-8 data in a member's profile fields or a very large payload). The current code passes the return value directly into the string:
  ```php
  $js = 'if ( typeof pmpromd_vars !== "undefined" ) { pmpromd_vars.marker_data = '
      . wp_json_encode( $all_marker_data ) . '; }';
  ```
  This injects `pmpromd_vars.marker_data = false;` into the page. The JS at `map.js:44` then tries to iterate an array that is now `false`, and `false.length` is `undefined` — no crash, but zero markers render with no indication of why. Add a guard:
  ```php
  $encoded = wp_json_encode( $all_marker_data );
  if ( empty( $encoded ) ) {
      return;
  }
  $js = 'if ( typeof pmpromd_vars !== "undefined" ) { pmpromd_vars.marker_data = ' . $encoded . '; }';
  ```

- **Major** (documentation gap) — The snippet runs an unbounded query across `users`, `pmpro_memberships_users`, and three `usermeta` JOINs with no LIMIT, then serializes the entire result set as inline JSON on every page load. On a directory with 500+ members this can produce a multi-hundred-KB inline script and a slow DB round-trip on every page view. The docblock says nothing about this. Before publish, add a note along the lines of: *"This snippet is suitable for directories with a moderate number of members. For sites with hundreds of active members, consider restricting levels via the `level` shortcode attribute and testing page load size."* The cache logic helps on repeat views, but the first request (and any cache miss) still hits the DB fully.

- **Minor** `show-all-member-markers-on-map-when-no-search-is-active.php:1` — `link: TBD` in the file header. Resolve this before the snippet goes live.

- **Minor** `show-all-member-markers-on-map-when-no-search-is-active.php:44` — `$_REQUEST['ps']` includes POST body and cookie data. This is a URL search parameter; use `$_GET['ps']` to be explicit about the source and avoid edge cases on sites with unusual server configs.

- **Minor** `show-all-member-markers-on-map-when-no-search-is-active.php:66-68` — The `first_name` and `last_name` LEFT JOINs and SELECTs are unused. `pmpromd_generate_marker_data()` calls `get_userdata($member->ID)` to get the full user object and never reads `->first_name` or `->last_name` from the result row. Dropping those two JOINs reduces the query cost slightly.

---

**Looks Good**

The `$level_sql` sanitization via `preg_replace('/[^0-9,]/', '', ...)` is correct — strips everything except digits and commas, making the interpolation safe. The cache interaction correctly wraps the custom SQL with the plugin's own `pmpromd_get/set_cached_results` helpers, which means the results are invalidated on membership and profile changes automatically. The `pmpromd_map_id` hook correctly returns `$map_id` unchanged while capturing attributes as a side effect.

flintfromthebasement (changes requested):
**Follow-up review**

**PR: strangerstudios/pmpro-snippets-library#487 — Create show-all-member-markers-on-map-when-no-search-is-active.php**
*dwanjuki → dev*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/487>

**Summary**

The map-markers snippet cleared both blocking issues from my last pass, but this update also bundles in two unrelated new snippets, and one of them has a critical bug: the Paystack batch-pagination is dead code, so any site with more than 25 active Paystack subscriptions will never have subscriptions beyond the first batch checked, forever. Still blocked, now for a different reason.

**Prior issues — status**

- **Resolved** `show-all-member-markers-on-map-when-no-search-is-active.php:104` — `wp_json_encode` failure is now guarded (`empty( $encoded )` → early return) before it hits `wp_add_inline_script`.
- **Resolved** (documentation gap) — Scale-warning paragraph added to the docblock almost verbatim as suggested.
- **Not addressed** `show-all-member-markers-on-map-when-no-search-is-active.php:1` — `link: TBD` is still in the file header; not touched by this diff.
- **Resolved** `:44` — `$_REQUEST['ps']` → `$_GET['ps']`.
- **Resolved** `:66-68` — unused `first_name`/`last_name` JOINs and columns removed.

New nit in the same file: the added guard mixes indentation —
```php
	if ( empty( $encoded ) ) {
	    return;
	}
```
the `return;` line uses a tab + 4 spaces instead of a tab. Minor, but WPCS will flag it.

**New findings (this diff)**

- **Critical** `send-billing-failure-emails-stale-paystack-subscriptions.php:38-46` — `PMPro_Subscription::get_subscriptions()` does not support an `offset` argument at all (verified against `classes/class-pmpro-subscription.php` in `paid-memberships-pro` — the method only reads `id`, `user_id`, `membership_level_id`, `status`, `subscription_transaction_id`, `gateway`, `gateway_environment`, `billing_amount`, `cycle_number`, `cycle_period`, `billing_limit`, `trial_amount`, `orderby`, and `limit`; there is no `offset` handling anywhere in the file). The `'offset' => $offset` key you pass is silently ignored. Every hourly run re-fetches the identical first `MY_PMPRO_PAYSTACK_BATCH_SIZE` (25) subscriptions ordered by `startdate DESC`, and `MY_PMPRO_PAYSTACK_OFFSET_KEY` just counts up in the background doing nothing. Any site with more than 25 active Paystack subscriptions will never have the remainder checked — the entire "batch through all subscriptions" premise the docblock describes doesn't work. Fix: either add real cursor-based pagination (e.g. `WHERE id > %d ORDER BY id ASC` using the last-processed subscription ID, since `get_subscriptions()` does support filtering by `id`), or drop batching and process everything in one pass if volume doesn't warrant it.

- **Major** `send-billing-failure-emails-stale-paystack-subscriptions.php:24` — `$secret_key = get_option( 'pmpro_paystack_lsk' )` always reads the *live* secret key. The gateway itself (`classes/class.pmprogateway_paystack.php` in `paystack-gateway-paid-memberships-pro`) switches between `pmpro_paystack_tsk` (test) and `pmpro_paystack_lsk` (live) based on `get_option( 'pmpro_gateway_environment' ) === 'sandbox'`. On any site running Paystack in test/sandbox mode, `$secret_key` is empty, the function returns on line 26 every single hour, and nothing is ever logged — a silent, permanent no-op for anyone testing this in sandbox before going live. Mirror the gateway's environment check.

- **Minor** `send-billing-failure-emails-stale-paystack-subscriptions.php:82-96` — 25 sequential `wp_remote_get()` calls per run with no delay between them. The docblock claims this design "stays well within Paystack API rate limits," but batching alone doesn't throttle — all 25 requests still fire back-to-back within the same execution. Not blocking, but the rate-limit claim in the docblock is currently untrue.

- **Minor** `downgrade-to-fallback-level-on-payment-failure.php:75` — `pmpro_changeMembershipLevel( $fallback_id, $user_id )` return value isn't checked in the failure path (the restore path on line 84 does check it). If the downgrade call fails, the code still proceeds as if it succeeded and removes the `pmpro_cancel_previous_subscriptions` filter with no indication anything went wrong. Low risk since a failed downgrade leaves the user at their original level (so restore later becomes a harmless no-op), but worth a `pmpro_log`/error_log for visibility.

**Looks Good**

The Paystack snippet correctly confirms status against the live API rather than trusting local state, guards against duplicate emails via a per-subscription user-meta flag, and clears that flag on the next successful payment — reasonable data-integrity approach apart from the pagination bug. The downgrade/restore snippet's use of `pmpro_cancel_previous_subscriptions` add/remove around `pmpro_changeMembershipLevel()` to preserve the retrying subscription is correct and matches the pattern used in PMPro core's own member-edit panel.

**Questions**

1. The Paystack and downgrade-to-fallback snippets are unrelated to this PR's original subject (the member-directory map). Is this intentionally a batch PR adding multiple recipes, or did a rebase pull in commits that belong elsewhere?
2. For the downgrade snippet: does `pmpro_getMembershipLevelForUser()` behave correctly for accounts with multiple simultaneous membership levels, or is this recipe scoped to single-level sites only? Worth a doc note either way.

PR #488, opened 2026-06-29
Add Paystack stale subscription billing failure email snippet

MaryOJob:
This PR adds a new `pmpro-paystack` folder under the `add-ons` directory to organize Paystack-specific recipes.

Inside, it includes a new snippet that runs hourly via PMPro's built-in Action Scheduler event (`pmpro_schedule_hourly`). It processes subscriptions in small batches to stay within Paystack API rate limits, checking the Paystack API as the source of truth to send billing failure emails for stale Paystack subscriptions.

### DIFF
--- a/add-ons/pmpro-paystack/send-billing-failure-emails-stale-paystack-subscriptions.php
+++ b/add-ons/pmpro-paystack/send-billing-failure-emails-stale-paystack-subscriptions.php
@@ -1,0 +1,163 @@
+<?php
+/**
+ * Send billing failure emails for stale Paystack subscriptions.
+ * 
+ * Runs hourly via PMPro's built-in Action Scheduler event (pmpro_schedule_hourly),
+ * processing subscriptions in small batches to stay well within Paystack API rate
+ * limits. Checks the Paystack API as the source of truth — only acts on 'attention'
+ * or 'cancelled' statuses. Sends billing failure emails once per subscription and
+ * flags it to prevent repeats.
+ *
+ * title: Send Billing Failure Emails for Stale Paystack Subscriptions
+ * layout: snippet
+ * collection: gateway, paystack
+ * category: billing, email
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+define( 'MY_PMPRO_PAYSTACK_BATCH_SIZE', 25 );
+define( 'MY_PMPRO_PAYSTACK_OFFSET_KEY', 'my_pmpro_paystack_stale_check_offset' );
+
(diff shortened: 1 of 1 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/pmpro-snippets-library#488 — Add Paystack stale subscription billing failure email snippet**
*MaryOJob → dev | 1 file, +163 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/488>

---

**Summary**

Good concept with a well-thought-out batching approach, but there's a logic bug that will permanently silence notifications for any subscription with no retrievable order — and the file header makes a promise the code doesn't keep. Not ready to merge as-is.

---

**Issues**

- **[Critical]** `send-billing-failure-emails-stale-paystack-subscriptions.php:149–155` — The `update_user_meta` stale-flag call is outside the `if ( ! empty( $order ) )` block:

  ```php
  if ( ! empty( $order ) ) {
      $member_email->sendBillingFailureEmail( ... );
      $admin_email->sendBillingFailureAdminEmail( ... );
  }
  // Flag this subscription so the notification is never sent again.
  update_user_meta( $subscription->get_user_id(), $flag_key, current_time( 'mysql' ) );
  ```

  When `get_orders(['limit' => 1])` returns nothing, no email is sent, but the subscription is flagged as notified and will never be processed again. Move `update_user_meta` inside the `if ( ! empty( $order ) )` block.

- **[Major]** `send-billing-failure-emails-stale-paystack-subscriptions.php:29–35` — The docblock for `my_pmpro_paystack_check_stale_subscriptions` is orphaned. It appears before the `my_pmpro_paystack_clear_stale_flag` function and its own docblock, leaving the hourly-check docblock attached to nothing. The two functions need to be reordered so each docblock is adjacent to the function it describes.

- **[Major]** `send-billing-failure-emails-stale-paystack-subscriptions.php:7` vs `:123` — The file-level docblock states the snippet "only acts on 'attention' or 'cancelled' statuses," but the actual `in_array` check contains only `'attention'`. Either `'cancelled'` is missing from the array, or the docblock needs correcting. Given that Paystack's `cancelled` status indicates a definitively failed subscription, intentionally omitting it should be a deliberate, documented decision — right now it looks like a mistake.

- **[Minor]** `send-billing-failure-emails-stale-paystack-subscriptions.php:81` — `current_time( 'timestamp', true )` is deprecated since WordPress 5.3. Use `time()` for a UTC timestamp.

- **[Minor]** `send-billing-failure-emails-stale-paystack-subscriptions.php:23–24` — `define()` at file scope without `if ( ! defined(...) )` guards. Low risk in practice, but Code Snippets can load files more than once in edge cases — add the guards.

---

**Verify Before Merge**

`pmpro_paystack_lsk` at line 64 — the option name used to fetch the secret key. The Paystack gateway repo isn't publicly accessible for verification. Confirm this matches the exact option name stored by the gateway, otherwise `$secret_key` will be empty every run and the function bails silently with no indication anything ran.

---

**Looks Good**

The offset-based batching logic is correct — resetting on empty batch and on partial batch, advancing otherwise. The early-exit guards (`pmpro_changeMembershipLevel`, `PMPro_Subscription`, secret key) are in the right place. `rawurlencode` on the subscription code before URL construction is correct.

PR #489, opened 2026-06-29
Add snippet to delay course and lesson access until launch date with countdown

MaryOJob:
This PR introduces a new recipe that restricts member access to courses and individual lessons until a designated launch date is reached. 

Key functionality included:
- Restricts non-eligible member access to PMPro Courses module post types before the launch date.
- Displays a dynamic countdown timer for members on restricted content pages.
- Leverages Paid Memberships Pro core functions to seamlessly match restriction settings with appropriate user membership levels.

### DIFF
--- a/add-ons/pmpro-courses/delay-course-access-until-launch-date.php
+++ b/add-ons/pmpro-courses/delay-course-access-until-launch-date.php
@@ -1,0 +1,109 @@
+<?php
+/**
+ * Delay Course and Lesson Access Until a Launch Date with Countdown for Members
+ * Blocks access to PMPro Courses Add On course and lesson content until a set launch
+ * date. Members who have purchased but cannot yet access will see a countdown card
+ * above the default restricted content message.
+ *
+ * title: Delay Course and Lesson Access Until a Launch Date with Countdown for Members
+ * layout: snippet
+ * collection: add-ons/pmpro-courses
+ * category: access
+ * link: tbd
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+/**
+ * Block access to course and lesson content until the launch date.
+ * Update $launch_date to your course launch date.
+ * Remove this snippet once the course has launched.
+ */
+function my_pmpro_delay_course_access_until_launch( $hasaccess, $post, $user, $levels ) {
+	// Set your course launch date (YYYY-MM-DD).
+	$launch_date = '2026-07-01';
+
+	// Only act before the launch date.
+	if ( current_time( 'Y-m-d' ) >= $launch_date ) {
+		return $hasaccess;
+	}
+
+	// Block access to course pages and individual lessons only.
+	if ( in_array( get_post_type( $post ), array( 'pmpro_course', 'pmpro_lesson' ), true ) ) {
+		return false;
+	}
+
+	return $hasaccess;
+}
+add_filter( 'pmpro_has_membership_access_filter', 'my_pmpro_delay_course_access_until_launch', 10, 4 );
+
+/**
+ * Show a countdown card above the restricted content message for members who have
+ * purchased but cannot access the course yet due to the launch date.
+ * Update $launch_date to match the date set in the access filter above.
+ * Remove this snippet once the course has launched.
+ */
+function my_pmpro_course_launch_countdown_content( $content ) {
+	// Set your course launch date (YYYY-MM-DD) — must match the date in the filter above.
+	$launch_date = '2026-07-01';
+
+	// Only act before the launch date.
+	if ( current_time( 'Y-m-d' ) >= $launch_date ) {
+		return $content;
+	}
+
+	// Only on course or lesson post types.
+	if ( ! is_singular( array( 'pmpro_course', 'pmpro_lesson' ) ) ) {
+		return $content;
+	}
+
+	// Only for logged-in members with a qualifying membership level.
+	if ( ! is_user_logged_in() || ! pmpro_hasMembershipLevel() ) {
+		return $content;
+	}
+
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#489 — Add snippet to delay course and lesson access until launch date with countdown**
*MaryOJob → dev | 1 file, +109 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/489>

**Summary**
Solid recipe overall — the access-block filter is scoped correctly and won't leak into unrelated post types or bypass the existing admin/free-preview logic in pmpro-courses. But the countdown message shows to any logged-in member with *any* membership level, not specifically members who can actually access this course, which contradicts the PR description's claim of matching "appropriate user membership levels." Worth a fix before merge; not blocking otherwise.

**Issues**

- **Major** `add-ons/pmpro-courses/delay-course-access-until-launch-date.php:70` — `my_pmpro_course_launch_countdown_content()` gates the countdown card on `pmpro_hasMembershipLevel()` with no arguments, which just checks "does this user have *any* membership level" (`includes/functions.php:1053` in core — empty `$levels` returns `! empty( $membership_levels )`). It doesn't check whether the user's level actually grants access to *this* course/lesson. A member with an unrelated level that has zero access to this course will see "You're registered! This course launches on..." — which is false, and they'll still be blocked after the launch date once your filter stops forcing `false`. You can't just call `pmpro_has_membership_access()` here since your own filter would force it to `false` before launch and recurse into the same wrong answer — the pattern already used elsewhere in pmpro-courses for exactly this problem (`includes/modules/tutorlms.php:177-181`, `includes/modules/senseilms.php:167-170`) is to `remove_filter()` your own callback, call `pmpro_has_membership_access( $post->ID, null, true )` to get the real levels/access, then `add_filter()` again. Use that pattern instead of the blanket `pmpro_hasMembershipLevel()` check.

- **Minor** `add-ons/pmpro-courses/delay-course-access-until-launch-date.php:20,55` — `$launch_date` is hardcoded independently in both functions with a comment telling the site owner to keep them in sync manually. One typo and the block date and the countdown date silently diverge. Define it once (e.g. a constant near the top of the file, or a shared helper) and reference it from both functions.

- **Minor** `add-ons/pmpro-courses/delay-course-access-until-launch-date.php:64` and the JS countdown target — the server-side gate uses `current_time( 'Y-m-d' )` (WP's configured timezone), but the JS does `new Date('2026-07-01T00:00:00')`, which the browser parses in the *visitor's local* timezone. A visitor in a timezone behind the site's will see the countdown hit zero and hide itself before the server actually unblocks the content (or vice versa for a visitor ahead). Worth a one-line note in the snippet header, or building the ISO timestamp with the site's UTC offset.

- **Minor** `add-ons/pmpro-courses/delay-course-access-until-launch-date.php:109` — file is missing a trailing newline (`\ No newline at end of file`); every other file in `add-ons/pmpro-courses/` ends with one.

**Looks Good**
The access-blocking filter (`my_pmpro_delay_course_access_until_launch`) is scoped tightly to `pmpro_course`/`pmpro_lesson` post types and runs on the generic `pmpro_has_membership_access_filter` hook before the type-specific one, so it doesn't interfere with the existing "Free/Public" lesson bypass (`pmpro_lessons_bypass_check` in pmpro-courses core) or the `manage_options` admin bypass in the redirect logic — both still work correctly ahead of launch.

**Questions**
1. Is the countdown meant to show for members of *any* level, or only members who purchased/qualify for this specific course? The PR description says the latter, but the code currently does the former.

flintfromthebasement (commented):
**Follow-up review**

**PR: strangerstudios/pmpro-snippets-library#489 — Add snippet to delay course and lesson access until launch date with countdown**
*MaryOJob → dev | 1 file, +131 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/489>

**Summary**
The blocking issue is fixed and fixed correctly. This clears my hold — good to merge once the trailing newline is added (still not there despite the commit message).

**Prior issues — status**

- **Resolved** — `delay-course-access-until-launch-date.php:70` (prior line numbers). The countdown now correctly gates on real course access, not "any membership level." `my_pmpro_course_launch_countdown_content()` temporarily `remove_filter()`s its own `my_pmpro_delay_course_access_until_launch` callback (matching priority 10), calls `pmpro_has_membership_access( $post->ID )` to get the user's actual access sans the launch-date block, then `add_filter()`s again — exactly the pattern I pointed to in `tutorlms.php`/`senseilms.php`. Verified `pmpro_has_membership_access()`'s signature in `includes/content.php:5` — `$post->ID` alone is correct, defaults to current user.

- **Resolved** — hardcoded `$launch_date` duplication. Both functions now read from a single `my_pmpro_get_course_launch_date()` helper. No more risk of the two dates drifting apart.

- **Partially addressed** — timezone mismatch. The JS now parses the launch date as UTC (`T00:00:00Z`) instead of the visitor's local time, so every visitor now sees the *same* countdown target — that's a real improvement over the old "different browser, different countdown" bug. But the PHP side still gates on `current_time( 'Y-m-d' )`, which resolves in the site's *configured* WP timezone, not UTC. Unless the site's timezone setting is UTC, there's still a fixed offset between when the countdown hits zero (UTC midnight) and when the server actually unblocks the content (site-timezone midnight). Better than before, not fully matched — a one-line comment noting "set your WP timezone to UTC, or adjust the JS offset to match" would close this out.

- **Not addressed** — `delay-course-access-until-launch-date.php:131` — the file is still missing a trailing newline (diff still shows `\ No newline at end of file` on the last line, `add_filter( 'the_content', ... );`). The commit message claims this was added ("added the missing trailing newline") but it wasn't — worth a quick follow-up commit since it's a one-character fix.

**New in this diff**
No new issues. The refactor is clean and doesn't touch anything outside the two functions it was meant to fix.

**Looks Good**
The remove/re-add filter dance is exactly right and mirrors the existing core pattern rather than inventing a new one — confirmed against `includes/content.php:135` that `pmpro_has_membership_access_filter` runs generically before the post-type-specific filter, so removing just this callback and calling the public function still exercises every other filter (admin bypass, free/public lesson bypass, etc.) correctly.

PR #490, opened 2026-06-29
Add snippet to downgrade member to fallback level on payment failure and restore on success

MaryOJob:
This PR introduces a new recipe that dynamically manages member levels during recurring billing disruptions.

### DIFF
--- a/membership-levels/downgrade-to-fallback-level-on-payment-failure.php
+++ b/membership-levels/downgrade-to-fallback-level-on-payment-failure.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Downgrade a member to a fallback level when a recurring payment fails,
+ * then restore their original level when a later payment succeeds.
+ *
+ * title: Downgrade member to fallback level on failed payment and restore on success
+ * layout: snippet
+ * collection: membership-levels
+ * category: subscription-handling
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+// Level to assign when payment fails. Use 0 to cancel membership.
+define( 'MY_PMPRO_FALLBACK_LEVEL_ID', 1 );
+
+// Stores the member's original level ID.
+define( 'MY_PMPRO_ORIGINAL_LEVEL_META_KEY', 'my_pmpro_pre_failure_level_id' );
+
+function my_pmpro_downgrade_on_payment_failure( $order ) {
+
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#490 — Add snippet to downgrade member to fallback level on failed payment and restore on success**
*MaryOJob → dev | 1 file, +80 -0 lines*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/490>

---

**Summary**
Solid concept and the hook pairing (`pmpro_subscription_payment_failed` / `pmpro_subscription_payment_completed`) is correct. The `pmpro_cancel_previous_subscriptions` filter suppression during downgrade is the right approach to keep retry subscriptions alive. One data integrity bug in the restore path needs fixing before merge; a few minor issues noted below.

---

**Issues**

**Major** `downgrade-to-fallback-level-on-payment-failure.php:73-79` — `delete_user_meta` is called unconditionally after `pmpro_changeMembershipLevel`. If the level change fails (invalid `$original_id`, DB error, level deleted since the downgrade), `pmpro_changeMembershipLevel` returns `false`, but the meta is still deleted. The user stays permanently on the fallback level with no restore state. Fix:

```php
$changed = pmpro_changeMembershipLevel( $original_id, $user_id );
if ( false !== $changed ) {
    delete_user_meta( $user_id, MY_PMPRO_ORIGINAL_LEVEL_META_KEY );
}
```

Use `!== false` not `=== true` — when the user already has the target level, `pmpro_changeMembershipLevel` returns `null` via a bare `return`, which is not a failure.

**Minor** `downgrade-to-fallback-level-on-payment-failure.php:17-19` — Both `define()` calls lack `if ( ! defined() )` guards. Re-inclusion or a constant name collision in another plugin causes a PHP fatal. Wrap each one.

**Minor** `downgrade-to-fallback-level-on-payment-failure.php:35-38` — The "avoid repeatedly downgrading" guard checks `$current_level->id === $fallback_id`. If the user sets `MY_PMPRO_FALLBACK_LEVEL_ID = 0` (the "cancel membership" use case mentioned in the comment), this check is never true (PMPro level IDs are always positive integers), so the downgrade action fires on every subsequent failure event instead of no-oping. Either update the comment to remove the `0` suggestion, or add an explicit early return when `$fallback_id === 0` and `empty( $current_level )`.

**Minor** `downgrade-to-fallback-level-on-payment-failure.php:80` — Missing newline at end of file.

---

**Looks Good**
- The `add_filter` / `remove_filter` bracket around `pmpro_changeMembershipLevel` in the downgrade path is clean and correct — it keeps the retry subscription alive without leaving a stale global filter.
- Storing the original level from `$order->membership_id` (the subscription's level) rather than re-querying the user's current level is the right source to use here.
- The symmetric meta cleanup in the restore path (delete after successful restore) is correct in intent; just needs the return-value guard.

PR #491, opened 2026-07-01
Prorate Seat Changes at Checkout for Group Accounts Parent Members

MaryOJob:
Adds a snippet that handles mid-cycle seat changes at checkout for existing 
Group Accounts parent members on Variable-seat, recurring levels.

- Upgrades: charges a prorated initial payment for additional seats only, 
  based on days remaining in the current billing cycle.
- Downgrades: charges $0 today and defers the seat count reduction to the 
  next renewal via user meta, so the parent retains access to already-paid 
  seats for the remainder of the cycle.
- Billing anchor date is preserved in both cases.
- Includes a checkout notice pre-filling the current seat count and 
  explaining the proration behavior to the parent member.
- Includes a live AJAX cost preview that updates as the seat count changes.
- No effect on Fixed-seat group levels or non-recurring parent levels.

### DIFF
--- a/add-ons/pmpro-group-accounts/prorate-group-accounts-seat-changes-at-checkout.php
+++ b/add-ons/pmpro-group-accounts/prorate-group-accounts-seat-changes-at-checkout.php
@@ -1,0 +1,375 @@
+<?php
+/**
+ * Prorate seat changes at checkout for existing Group Accounts parent members.
+ *
+ * For parent members on Variable-seat, recurring Group Accounts levels, this snippet
+ * adjusts the checkout flow so that seat changes are handled fairly mid-cycle. When
+ * adding seats, only a prorated amount is charged for the remaining days in the current
+ * billing cycle. When removing seats, nothing is charged today — the seat count reduction
+ * is deferred until the next renewal so the parent retains access to the seats they
+ * already paid for. In both cases, the existing subscription billing date is preserved.
+ *
+ * This snippet has no effect on Fixed-seat group levels or non-recurring parent levels.
+ *
+ * title: Prorate Seat Changes at Checkout for Group Accounts Parent Members
+ * layout: snippet
+ * collection: add-ons, group-accounts
+ * category: group-accounts, billing, checkout
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+/**
+ * For existing group parents on Variable-seat, recurring levels, show current seat count
+ * and pre-fill the seats field. Runs before the group-accounts plugin's own
+ * pmpro_checkout_boxes hook (priority 10).
+ *
+ * Fixed-seat groups and non-recurring levels are skipped entirely.
+ */
+function my_pmpro_group_parent_existing_seats_notice() {
+	if ( ! function_exists( 'pmprogroupacct_get_settings_for_level' ) || ! class_exists( 'PMProGroupAcct_Group' ) ) {
+		return;
+	}
+
+	$user_id = get_current_user_id();
+	if ( ! $user_id ) {
+		return;
+	}
+
+	$level = pmpro_getLevelAtCheckout();
+	if ( empty( $level->id ) ) {
+		return;
+	}
+
+	$settings = pmprogroupacct_get_settings_for_level( $level->id );
+	if ( empty( $settings ) ) {
+		return;
+	}
+
+	// Only show for Variable-seat groups. Fixed groups have no editable #pmprogroupacct_seats
+	// field at checkout, so showing "enter the new total you want" would be misleading.
+	// Variable groups are identified by min_seats !== max_seats; Fixed groups lock both to
+	// the same value since there is no range to choose from.
+	if ( ! isset( $settings['min_seats'], $settings['max_seats'] ) || (int) $settings['min_seats'] === (int) $settings['max_seats'] ) {
+		return;
+	}
+
+	// Only show for recurring levels. Non-recurring levels charge the full cost of all
+	// seats as a fresh purchase — proration does not apply and the notice would be misleading.
+	if ( ! pmpro_isLevelRecurring( $level ) ) {
+		return;
+	}
+
+	$existing_group = PMProGroupAcct_Group::get_group_by_parent_user_id_and_parent_level_id( $user_id, $level->id );
+	if ( empty( $existing_group ) ) {
+		return;
+	}
+
+	$existing_seats = intval( $existing_group->group_total_seats );
+	$ajax_url       = admin_url( 'admin-ajax.php' );
+	$ajax_nonce     = wp_create_nonce( 'my_pmpro_seats_cost_text' );
+	?>
+	<script>
+	document.addEventListener( 'DOMContentLoaded', function () {
+		var field = document.getElementById( 'pmprogroupacct_seats' );
+		if ( ! field ) {
+			return;
+		}
+		// Pre-fill with existing seat count.
+		field.value = <?php echo (int) $existing_seats; ?>;
+
+		// Live-update the level cost text whenever the seat count changes.
+		var levelId    = <?php echo (int) $level->id; ?>;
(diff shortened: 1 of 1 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/pmpro-snippets-library#491 — Prorate Seat Changes at Checkout for Group Accounts Parent Members**
*MaryOJob → dev | 1 file, +375 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/491>

**Summary**
Not ready to merge. The proration math and downgrade-deferral concept are sound, but three critical defects break the core promise of this snippet: the deferred downgrade never actually applies at renewal, the live cost-preview AJAX call has unintended side effects on billing state, and the initial-payment override silently drops legitimate charges. Verified all three against the current `paid-memberships-pro` and `pmpro-group-accounts` source.

**Issues**

- **Critical** `prorate-group-accounts-seat-changes-at-checkout.php:349-353` — `my_pmpro_apply_pending_seat_reduction()` is hooked to `pmpro_subscription_payment_completed` and gated on `is_a( $subscription, 'PMPro_Subscription' )`. Every fire site of this action across core and gateway add-ons (`includes/gateway-request-handlers.php:203`, `services/braintree-webhook.php:177`, `services/authnet-silent-post.php:151`, `services/twocheckout-ins.php:109`, `pmpro-payfast/services/payfast_itn_handler.php:422`) passes a `MemberOrder` object, never a `PMPro_Subscription`. `MemberOrder` does not extend `PMPro_Subscription`, so the guard is always false and the function returns immediately — the deferred seat reduction is never applied. Seats get reverted to the pre-downgrade count at checkout (line 332-338) but the reduction promised "at the next renewal" never happens. Members who downgrade keep the higher seat count forever. Fix: derive the subscription from `$order->user_id` / `$order->membership_id` (e.g. via `PMPro_Subscription::get_subscriptions_for_user()`), don't gate on the wrong class.

- **Critical** `prorate-group-accounts-seat-changes-at-checkout.php:282-291` — `my_pmpro_prorate_parent_seat_upgrade()` unconditionally calls `update_user_meta()`/`delete_user_meta()` on `_my_pmpro_pending_seat_reduction` every time it runs. It's hooked to `pmpro_checkout_level` (line 295), which is also invoked by `my_pmpro_seats_cost_text_ajax()` at line 180 via `pmpro_getLevelAtCheckout( $level_id )` — i.e. on every debounced keystroke of the live cost preview (line 129), before the user submits anything. This means: (a) a mere price preview at a lower seat count silently writes a pending-downgrade record for an order that was never placed, and (b) a preview at a higher/equal seat count silently deletes a legitimate pending downgrade the user scheduled earlier, canceling it without their knowledge. A read-only preview must not mutate persistent billing state. Fix: only touch `_my_pmpro_pending_seat_reduction` from the actual checkout submission path (e.g. gate on `pmpro_after_checkout`/order context), not from every `pmpro_checkout_level` invocation.

- **Critical** `prorate-group-accounts-seat-changes-at-checkout.php:267` — `$level->initial_payment = $additional_seats * $price_per_seat * $proration_ratio;` is a straight overwrite, not a delta adjustment. By the time this filter runs (priority 20), `pmprogroupacct_pmpro_checkout_level_parent()` (priority 10, `includes/parents.php:243`/`254`) has already done `$level->initial_payment += $seat_cost` where `$seat_cost` covers the *full new seat total*, on top of whatever base `initial_payment` the admin configured on the level itself (a signup fee independent of per-seat pricing is a valid Group Accounts config). This line discards both the base fee and the group-accounts seat surcharge, replacing them with only the prorated additional-seat amount. Any level with a nonzero base `initial_payment` silently loses that charge on every seat-adjustment checkout. Fix: subtract what group-accounts added (`$new_total_seats * $price_per_seat`) before adding back the prorated additional-seat charge, rather than reassigning outright.

- **Minor** `prorate-group-accounts-seat-changes-at-checkout.php:267` — no `round()` on the computed `initial_payment`; the `proration_ratio` division can leave many decimal places, which some gateways reject. Round to 2 decimals before assigning.

- **Minor** `prorate-group-accounts-seat-changes-at-checkout.php:332-338,370-376` — reconstructs `$wpdb->prefix . 'pmprogroupacct_groups'` instead of the `$wpdb->pmprogroupacct_groups` property the plugin itself registers (`pmpro-group-accounts.php:36`). Functionally identical here, but inconsistent with the codebase convention.

- **Minor** — `_my_pmpro_pending_seat_reduction` is a single-valued user meta key, not scoped per group/level. A parent managing seat changes on two different variable-seat parent levels concurrently would have the second downgrade clobber the first's pending state.

**Looks Good**
The seat-tier detection (`min_seats !== max_seats`) and priority ordering against `pmpro-group-accounts`' own hooks (5/10/20/999) are correctly reasoned and match the actual plugin hook priorities I checked in `parents.php`.

**Questions**
1. Is a base `initial_payment` on a Variable-seat parent level actually a supported/expected config in practice, or is `initial_payment` always 0 for these levels in your deployments? That changes how urgent finding #3 is.
2. Was the AJAX preview's side effect on `_my_pmpro_pending_seat_reduction` (finding #2) intentional, or was `pmpro_checkout_level` assumed to only fire on real submission?

flintfromthebasement (approved):
**Follow-up review**

**PR: strangerstudios/pmpro-snippets-library#491 — Prorate Seat Changes at Checkout for Group Accounts Parent Members**
*MaryOJob → dev | 1 file, +419 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/491>

**Summary**
All three Critical issues from my prior review are resolved and verified against current core (`paid-memberships-pro`) and `pmpro-group-accounts` source. No new issues introduced. Ready to merge.

**Prior issues — status**

- **Resolved** `prorate-group-accounts-seat-changes-at-checkout.php:378-388` — `my_pmpro_apply_pending_seat_reduction()` now type-checks `MemberOrder` (not `PMPro_Subscription`) and derives `$user_id`/`$level_id` from `$order->user_id`/`$order->membership_id`. Verified `MemberOrder` defines both properties (`classes/class.memberorder.php:798-799`) and every fire site of `pmpro_subscription_payment_completed` in core (`gateway-request-handlers.php:203`, `braintree-webhook.php:177`, `authnet-silent-post.php:151`, `twocheckout-ins.php:109`) plus `pmpro-payfast/services/payfast_itn_handler.php:422` passes a `MemberOrder`. The deferred downgrade will now actually apply at renewal.

- **Resolved** `prorate-group-accounts-seat-changes-at-checkout.php:297-320` — the AJAX preview path is now excluded from meta writes via `$is_preview = $_REQUEST['action'] === 'my_pmpro_seats_cost_text'`. Confirmed the real checkout form (`preheaders/checkout.php:74` → `pmpro_getLevelAtCheckout()`) never sends an `action` request param, so the heuristic correctly distinguishes preview calls from real submissions. A price preview no longer writes or clears pending-downgrade state.

- **Resolved** `prorate-group-accounts-seat-changes-at-checkout.php:282-284` — `$level->initial_payment` is now adjusted by subtracting the exact Group Accounts surcharge (`$new_total_seats * $price_per_seat`) before adding the prorated charge, instead of overwriting. Verified this matches `pmprogroupacct_pmpro_checkout_level_parent()` exactly: `$level->initial_payment += $seat_cost` where `$seat_cost = $seats * (float)$settings['pricing_model_settings']` (`pmpro-group-accounts/includes/parents.php:231,254`). Any base signup fee on the level is now preserved.

- **Resolved** `prorate-group-accounts-seat-changes-at-checkout.php:267` — `round(..., 2)` added.

- **Resolved** `prorate-group-accounts-seat-changes-at-checkout.php:362,410` — both `$wpdb->update()` calls now use `$wpdb->pmprogroupacct_groups`, matching the property the plugin registers (`pmpro-group-accounts.php:36`).

- **Resolved** — meta key is now scoped per level (`_my_pmpro_pending_seat_reduction_{$level_id}`), so two concurrently-managed parent levels no longer clobber each other's pending state.

**New findings**
None. The `is_preview` detection is client-suppliable in principle (a tampered checkout POST could add `action=my_pmpro_seats_cost_text` to skip the meta write), but tracing the consequence through `my_pmpro_defer_seat_reduction_after_checkout()` shows it only results in the seat reduction applying immediately instead of at renewal — no financial or security exposure, so not worth blocking on.

**Looks Good**
Clean resolution of all three critical defects with correct cross-referencing against the actual `pmprogroupacct_pmpro_checkout_level_parent()` and `pmpro_subscription_payment_completed` call sites — this wasn't guessed, it was traced.

PR #492, opened 2026-07-07
Add UK VAT-Compliant Invoicing to PMPro with Stripe 

dwanjuki:
Adding recipe and some minor updates that don't change functionality.

Related trello: https://trello.com/c/sIzYbqAx

### DIFF
--- a/payment-gateways/stripe/create-uk-vat-compliant-invoices.php
+++ b/payment-gateways/stripe/create-uk-vat-compliant-invoices.php
@@ -22,6 +22,15 @@
  * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
  */
 
+// Company details used throughout this snippet's VAT invoicing output.
+if ( ! defined( 'MY_PMPRO_VAT_COMPANY_NAME' ) ) {
+	define( 'MY_PMPRO_VAT_COMPANY_NAME', 'Your Company Name' );
+	define( 'MY_PMPRO_VAT_COMPANY_ADDRESS', '123 Example Street, Anytown, AB1 2CD' );
+	define( 'MY_PMPRO_VAT_COMPANY_NUMBER', 'GB123456789' );
+	define( 'MY_PMPRO_VAT_COMPANY_EMAIL', 'billing@example.com' );
+	define( 'MY_PMPRO_VAT_COMPANY_URL', 'https://example.com' );
+}
+
 /**
  * Fetch tax IDs from Stripe and cache them in user meta.
  * Returns the cached value immediately if one already exists.
@@ -38,7 +47,7 @@ function my_pmpro_fetch_stripe_tax_ids( $user_id, $order ) {
 	// Return cached value if already fetched. 'none' means the user had no tax IDs at fetch time.
 	$vat_info = get_user_meta( $user_id, 'pmpro_stripe_tax_ids', true );
 	if ( ! empty( $vat_info ) ) {
-		return $vat_info === 'none' ? '' : $vat_info;
+		return 'none' === $vat_info ? '' : $vat_info;
 	}
 
 	try {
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#492 — Add UK VAT-Compliant Invoicing to PMPro with Stripe**
*dwanjuki → dev | 1 file, +295 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/492>

**Summary**
Single new recipe file, well-documented, all hooks/filters used (`pmpro_email_data`, `pmpro_invoice_bullets_bottom`, `pmpro_manage_orderslist_columns`, `pmpro_manage_orderlist_custom_column`, `pmpro_order_single_meta`, `pmpro_get_price_parts_with_total`, `pmpro_after_checkout`, `pmpro_updated_order`) verified to exist in PMPro core `dev`. One real markup bug to fix before merge; otherwise close to ready.

**Issues**

- **Major** `create-uk-vat-compliant-invoices.php:170-179` (`my_pmpro_invoice_show_tax_ids`) — `pmpro_invoice_bullets_bottom` fires inside `<ul>...</ul>` in `pages/invoice.php:226-244` (list expects `<li>` children). This callback echoes a `<div class="pmpro_vat_invoice">` containing a `<p>` directly inside that `<ul>`. That's invalid nesting — browsers will hoist the `<div>` out of the list, which can visually break the "more information" block depending on theme CSS. Wrap the content in an `<li>` (or at minimum avoid block-level elements as direct `<ul>` children) to match how other bullet-hook consumers render.

- **Minor** `create-uk-vat-compliant-invoices.php:206-224` (`my_pmpro_admin_order_show_tax_ids`) — mixed indentation: this function uses spaces where the rest of the file uses tabs. Doesn't affect behavior, but worth a pass through phpcbf before merge.

- **Minor** `create-uk-vat-compliant-invoices.php:132, 178, 234` — VAT invoice number is just the internal order ID cast to int (`(int) $morder->id`). That's fine as a recipe default, but flag in the doc comment that UK VAT invoicing technically expects a dedicated sequential/gapless invoice series — using the raw order ID works only if orders are never deleted/renumbered. One line in the header comment would save a support ticket later.

**Looks Good**
- Caching pattern in `my_pmpro_fetch_stripe_tax_ids` (sentinel `'none'` value to avoid re-hitting the Stripe API on every email/render) is correct and avoids the obvious perf trap.
- Priority-5 hook ordering (`pmpro_after_checkout`/`pmpro_updated_order` at priority 5, ahead of PMPro's own email send at 10) is the right way to guarantee user meta is populated before the template variables are needed — nice attention to hook ordering.
- Output escaping (`esc_html`, `esc_attr`) is consistently applied everywhere user-influenced data (VAT number from Stripe) is echoed.
- `gettext` filter is scoped tightly to the `paid-memberships-pro` domain and exact string matches, so it won't leak into unrelated translations.

**Questions**
1. The placeholder company details (`Your Company Name`, `GB123456789`, `example.com`) are hardcoded in three separate places (email filter, frontend invoice block, admin single-order meta). Worth pulling into one constant/function so a site owner only has to edit it once instead of three?

flintfromthebasement (commented):
**Follow-up review**

**PR: strangerstudios/pmpro-snippets-library#492 — Add UK VAT-Compliant Invoicing to PMPro with Stripe**
*dwanjuki → dev | 1 file, +295 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/492>

**Summary**
All three prior issues are addressed. This clears my blocking concern and is ready to merge — one small doc-comment suggestion remains optional.

**Prior issues — status**

- **Resolved** `create-uk-vat-compliant-invoices.php:170-179` (Major, invalid `<div>`/`<p>` inside `<ul>`) — now wrapped in `<li>...</li>` (diff lines 197/211). Matches list markup expected by `pages/invoice.php`.
- **Resolved** `create-uk-vat-compliant-invoices.php:206-224` (Minor, mixed indentation) — file now consistently uses tabs and Yoda conditionals (`'none' === $vat_info`, `'tax_ids' !== $column_name`) throughout.
- **Resolved** (Question) hardcoded company details in three places — pulled into `MY_PMPRO_VAT_COMPANY_*` constants defined once near the top of the file and reused in the email filter, frontend bullet block, and admin single-order meta. Also now consistently escaped (`esc_html`/`esc_url`/`esc_attr`) at each use site, including the admin meta block which wasn't escaped before.
- **Not addressed** VAT invoice number using the raw order ID (`(int) $morder->id`) as the invoice number — no doc-comment caveat added about gapless/sequential invoice numbering expectations. Still a one-line nice-to-have, not blocking.

**New findings (scoped to this diff)**

None. The `strip_tags` → `wp_strip_all_tags` swap and the Stripe API call reformatting are no-op style/lint cleanups (likely from phpcbf) with no behavioral change.

**Looks Good**
- Constant-guard pattern (`if ( ! defined(...) )`) for the company details means a site owner can override any of them from `wp-config.php` or another snippet without editing this file — good ergonomics for a shared recipe.
- Escaping is now applied uniformly across all three output surfaces (email, frontend invoice, admin order meta), closing the one gap that existed in admin meta previously.

Nothing else open. Approving direction-wise; the remaining item is optional polish.

PR #493, opened 2026-07-07
Add a Lock icon to protected post titles for members without access

dwanjuki:
A customer requested similar functionality.
Also related: https://www.paidmembershipspro.com/locked-post-message-image-overlay/

### DIFF
--- a/frontend-pages/add-lock-icon-to-protected-post-titles.php
+++ b/frontend-pages/add-lock-icon-to-protected-post-titles.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Add a lock icon before the title of posts that require membership,
+ * shown only to users who do not have access to the post.
+ *
+ * title: Add a Lock Icon to Protected Post Titles
+ * layout: snippet
+ * collection: frontend-pages
+ * category: content-protection
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+function my_pmpro_lock_icon_in_title( $title, $id = null ) {
+
+	// Skip in the admin, in feeds, and when no post ID is available.
+	if ( is_admin() || is_feed() || ! $id ) {
+		return $title;
+	}
+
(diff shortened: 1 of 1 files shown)


flintfromthebasement (approved):
**PR: strangerstudios/pmpro-snippets-library#493 — Add a Lock icon to protected post titles for members without access**
*dwanjuki → dev | 1 file, +49 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/493>

**Summary**
Clean, single-function snippet that prepends a lock icon to `the_title` for posts the current visitor can't access. Logic is correct, guards are appropriate, output is escaped. Ready to merge.

**Issues**

- **Minor** `frontend-pages/add-lock-icon-to-protected-post-titles.php:33` — `$access[1]` / `$access[0]` are magic indices into the array returned by `pmpro_has_membership_access( $id, null, true )` (verified against `includes/content.php:139` in core: `array($hasaccess, $post_membership_levels_ids, $post_membership_levels_names)`). Destructuring would read clearer and survive a future core signature reorder better: `list( $has_access, $level_ids ) = pmpro_has_membership_access( $id, null, true );` then `if ( ! empty( $level_ids ) && empty( $has_access ) )`.
- **Minor** `frontend-pages/add-lock-icon-to-protected-post-titles.php:24-38` — `pmpro_has_membership_access()` runs its own DB query with no caching, and this filter fires once per rendered title. On archive/search pages with many posts this is an extra query per post per page load. Matches the existing pattern in `restricting-content/add-locked-post-message-image-overlay-on-protected-content.php` (same repo, same tradeoff), so not a blocker — just flagging it's an N-query pattern if someone later asks "why is my archive page slow."

**Looks Good**
- Correct guard order: bails on admin/feed/missing ID before touching PMPro, then checks `in_the_loop()` before checking membership — avoids running the access check anywhere it wouldn't render (widgets, REST responses, `wp_title`).
- `access[1]` (levels required) and `access[0]` (has access) are read correctly — icon only shows when the post requires membership *and* the current user lacks access. Verified against `pmpro_has_membership_access()` in `includes/content.php`.
- Output properly escaped (`esc_url`, `esc_attr__`), and `PMPRO_URL . '/images/lock.svg'` resolves to a real asset in core (`images/lock.svg`).
- `function_exists`/`defined` guard for PMPro absence is correct here — this is a standalone snippet, not something with its own hook lifecycle to defer.

PR #495, opened 2026-07-21
Code recipe for sorting the Member Directory by user field.

RachelRVasquez:
The original gist was being hosted on a different account: https://gist.github.com/ipokkel/2f8ffeceefcdbeba764f69cb9d12304e#file-sort-pmpro-member-directory-order-by-custom-user-meta-field-php

We had an existing snippet in our library but the existing snippet was broken and caused an unknown database column error. I had Flint compare the two code snippets, and he made a suggested fix. I combined his suggested fix and a second PR review from Flint, with piece's from iPokkel's original gist, to get this final version. The doc block should meet standards and the function name should be clearer.

The code recipe has been tested locally, but could use another set of eyes from someone more familiar with the **Member Directory and Profile Pages Add On**.


flintfromthebasement (changes requested):
**PR: strangerstudios/pmpro-snippets-library#495 — Code recipe for sorting the Member Directory by user field.**
*RachelRVasquez → dev | 1 file, +41 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/495>

---

**Summary**
Don't merge as-is. This snippet overwrites the Member Directory's `SELECT` clause with an obsolete column list from an older version of the add-on, which will break map pins and drop columns the current template never asked for. It's also a near-duplicate of an existing recipe (`sort-pmpro-member-directory-order-by-custom-user-meta-field.php`) in the same folder with the identical bug. The fix is small — the author just needs to stop replacing `SELECT` and only touch `JOIN`/`ORDER`.

---

**Issues**

- **Critical** `order-pmpro-member-directory-by-custom-user-meta-field.php:19` — `$parts['SELECT']` is replaced wholesale with a hardcoded list that doesn't match what the current add-on builds. Verified against `repos/strangerstudios/pmpro-member-directory/templates/directory.php:84-92`: the real base `SELECT` only needs `u.ID, u.user_nicename, umf.meta_value AS first_name, uml.meta_value AS last_name, ummap.meta_value AS maplocation` — everything else the directory template needs (`display_name`, `user_email`, etc.) is re-fetched later via `get_userdata( $auser->ID )` (`directory.php:375,412`). The snippet's replacement drops `ummap.meta_value AS maplocation` entirely, so any site running Member Maps alongside this recipe gets pins silently missing for every member, and adds a pile of membership/billing columns (`mu.membership_id`, `mu.initial_payment`, etc.) the template doesn't read. This is copied straight from the old gist against a prior plugin architecture. **Fix: delete the `$parts['SELECT']` line.** It isn't needed — append the `umk` join and reference `umk.meta_value` directly in `ORDER`.

- **Critical** `order-pmpro-member-directory-by-custom-user-meta-field.php:36` — `ORDER BY user_meta_key_name` only works because that alias is defined in the replaced `SELECT` (line 19: `umk.meta_value as user_meta_key_name`). If the `SELECT` override is removed per the fix above, this breaks. **Fix together with the above: `$parts['ORDER'] = ' ORDER BY umk.meta_value ' . $order . ' ';`** — references the `JOIN` alias directly, no `SELECT` change needed.

- **Minor** `order-pmpro-member-directory-by-custom-user-meta-field.php:19` — `SQL_CALC_FOUND_ROWS` is dead weight. Confirmed the add-on computes total rows via a separate regex-derived `SELECT COUNT( DISTINCT u.ID )` query (`directory.php:167-181`), not `SQL_CALC_FOUND_ROWS`. Goes away automatically once the `SELECT` override is removed.

- **Major** (whole file) — Near-duplicate of the existing `add-ons/pmpro-member-directory/sort-pmpro-member-directory-order-by-custom-user-meta-field.php`, confirmed side by side: same filter (`pmpro_member_directory_sql_parts`), same broken `SELECT` replacement (including the same stray `SQL_CALC_FOUND_ROWS` and dropped `maplocation`), same `LEFT JOIN usermeta umk` pattern, same `ORDER BY user_meta_key_name` alias trick. Differences are cosmetic (function name, example field `company_name` vs `company`, a `$switch_sort_order` boolean vs a plain `$order` string). Two recipes for the same use case with near-identical filenames in the same collection will confuse anyone browsing the library, and the underlying bug now exists in two places instead of one. Either replace the old file with this one (once fixed) or explain in the PR why both need to exist.

- **Minor** `order-pmpro-member-directory-by-custom-user-meta-field.php` — Missing trailing newline (`\ No newline at end of file` in the diff).

---

**Looks Good**
- `esc_sql()` applied to the meta key name before interpolating into the `JOIN` clause — good defensive habit for a recipe people will copy-paste and edit.
- `$parts['JOIN'] .=` appends rather than clobbers the plugin's existing joins — correct pattern.
- Doc block front matter (`title`, `layout`, `collection`, `category`, `link`) matches library conventions, and the linked doc page exists.

---

**Questions**
1. Was the existing `sort-pmpro-member-directory-order-by-custom-user-meta-field.php` known about when this was written? If the intent is to supersede it, this PR should update/remove that file rather than leave two similar-but-buggy recipes in the library.
2. Once the `SELECT` override is dropped, should users with no value set for the custom field sort first or last? Worth a one-line note in the recipe since MySQL's default NULL ordering may not be what a site owner expects.

RachelRVasquez:
Responding to Flint's Review:

> * **Major** (whole file) — Near-duplicate of the existing `add-ons/pmpro-member-directory/sort-pmpro-member-directory-order-by-custom-user-meta-field.php`...

I've deleted my duplicate and updated the existing code recipe snippet instead.

> * **Minor** `order-pmpro-member-directory-by-custom-user-meta-field.php` — Missing trailing newline (`\ No newline at end of file` in the diff).

I've added the newline. ✅

RachelRVasquez:
Responding to Flint's first PR review continued. He did a second review in Slack where we could go back and forth with questions.

The first two critical issues have been addressed. The existing snippet that I originally thought was the same as the one I tried to add was not. The existing snippet was the one based on an old database structure and was causing database errors. Flint offered a fix which I tweaked further based on Flint's second PR review and iPokkel's original gist.

```
function pmpro_sort_member_directory_by_user_field( $parts ) {
    global $wpdb;
    $user_meta_key_name = 'company';
    $order = 'ASC';

    $parts['JOIN']  .= " LEFT JOIN $wpdb->usermeta umk ON umk.meta_key = '" . esc_sql( $user_meta_key_name ) . "' AND u.ID = umk.user_id ";
    $parts['ORDER']  = " ORDER BY umk.meta_value " . esc_sql( $order ) . " ";
    return $parts;
}
add_filter( 'pmpro_member_directory_sql_parts', 'pmpro_sort_member_directory_by_user_field', 10, 1 );
```

PR #496, opened 2026-08-03
Add send-invoice-email-after-membership-checkout.php

dwanjuki:
Migrating snippet to library: https://gist.github.com/travislima/8027b0f081506c7c3d68878296358750

Updated to resolve a fatal error: `Uncaught TypeError: PMPro_Email_Template_Invoice::__construct(): Argument #1 ($user) must be of type WP_User, string given`

### DIFF
--- a/email/send-invoice-email-after-membership-checkout.php
+++ b/email/send-invoice-email-after-membership-checkout.php
@@ -14,7 +14,7 @@
  * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
  */
 
-function pmpro_after_checkout_send_invoice_email( $user_id, $order ) {
+function my_pmpro_after_checkout_send_invoice_email( $user_id, $order ) {
 
 	$user = get_user_by( 'id', $user_id );
 
@@ -25,4 +25,4 @@ function pmpro_after_checkout_send_invoice_email( $user_id, $order ) {
 	$email = new PMProEmail();
 	$email->sendInvoiceEmail( $user, $order );
 }
-add_action( 'pmpro_after_checkout', 'pmpro_after_checkout_send_invoice_email', 10, 2 );
+add_action( 'pmpro_after_checkout', 'my_pmpro_after_checkout_send_invoice_email', 10, 2 );
--- a/email/send-invoice-email-after-membership-checkout.php
+++ b/email/send-invoice-email-after-membership-checkout.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * Send the Recurring Payment Receipt (Invoice) email on initial checkout.
+ *
+ * title: Send Members an Additional Invoice via Email after Membership Checkout
+ * layout: snippet
+ * collection: email
+ * category: emails
+ * link: https://www.paidmembershipspro.com/send-members-an-additional-invoice-via-email-after-membership-checkout/
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#496 — Add send-invoice-email-after-membership-checkout.php**
*dwanjuki → dev | 1 file, +28 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/496>

---

**Summary**
Functionally correct but should fire on the right event before merging. The type fix (`get_user_by('id', $user_id)` → `WP_User`) matches core's actual constructor signature (`PMPro_Email_Template_Invoice::__construct( WP_User $user, MemberOrder $order )`), and the hook usage matches how core fires `pmpro_after_checkout`. The real problem is that this hook fires before PMPro's own checkout-email gate, so the snippet ignores a site's decision to disable checkout emails.

---

**Issues**

- **Major** `email/send-invoice-email-after-membership-checkout.php:27` — `pmpro_after_checkout` fires at `includes/checkout.php:297`, immediately *before* core's own email gate at `includes/checkout.php:300`: `if ( apply_filters( 'pmpro_send_checkout_emails', true, $order ) ) { ... }`. Any site that disables checkout emails via that filter (staging, headless integrations, custom workflows) will still get this invoice email, since the snippet runs unconditionally on an action that fires ahead of the guard. Wrap the send:
  ```php
  if ( apply_filters( 'pmpro_send_checkout_emails', true, $order ) ) {
      $email = new PMProEmail();
      $email->sendInvoiceEmail( $user, $order );
  }
  ```

- **Minor** `email/send-invoice-email-after-membership-checkout.php:16` — Function name `pmpro_after_checkout_send_invoice_email` uses a bare `pmpro_` prefix. Every other snippet in `email/` uses a `my_pmpro_`-style prefix, and this specific name is stylistically close to real core functions hooked to the same action (`pmpro_after_checkout_save_fields`, `pmpro_after_checkout_update_consent`, `pmpro_after_checkout_reset_recaptcha`). Rename to `my_pmpro_after_checkout_send_invoice_email` to avoid confusion and reduce redeclaration risk if a customer's own code follows core naming.

- **Minor** `email/send-invoice-email-after-membership-checkout.php:25` — No guard for $0 orders. `sendInvoiceEmail()` (`classes/class.pmproemail.php:786-800`) only bails if `$user`, `$order`, or `$order->membership_id` is empty — there's no check on payment amount, so free-level checkouts get a "Recurring Payment Receipt" invoice for $0. Not a blocker, just confusing to members. Consider gating on `$order->initial_payment > 0`.

---

**Looks Good**
- The `get_user_by('id', $user_id)` + null check is the correct, verified fix for the reported `TypeError` — `PMPro_Email_Template_Invoice` genuinely requires a `WP_User`, and `get_user_by()` returns one (or `false`, which the guard catches).
- Hook name, priority, and accepted-args count match core's `do_action( "pmpro_after_checkout", $order->user_id, $order )` exactly.
- `sendInvoiceEmail()` has its own internal guards for missing user/order, so no redundant validation is needed in the snippet beyond the existing user check.

---

**Questions**

1. Should the send be wrapped in `pmpro_send_checkout_emails` per the Major finding above, or is bypassing that filter intentional (e.g., invoices are considered distinct from "checkout emails")? Worth confirming with the customer/gist author before merging.

dwanjuki:
@flintfromthebasement the `pmpro_send_checkout_emails` filter bypass is intentional, this invoice is considered distinct from the checkout confirmation email.

PR #501, opened 2026-08-12
Add Monthly/Annual Pricing Toggle for PMPro Levels Page and Advanced Levels Page

MaryOJob:
Adds a frontend pricing toggle (Monthly / Annual) to the PMPro Levels Page and 
Advanced Levels Page Add-on. Members can switch the displayed price between an annual figure and its monthly-equivalent breakdown without a page reload.

- Detects the levels shortcode/block via `has_shortcode()` (`pmpro_levels`, 
  `pmpro_advanced_levels`) and `has_block()` (`pmpro-advanced-levels/advanced-levels-page`) 
  — not tied to a specific page slug, so it works regardless of site configuration.
- Parses the rendered annual price text client-side (regex on currency symbol + 
  amount) and computes the monthly equivalent (`annual / 12`) in JS.
- Renders a toggle UI above the levels container, falling back to a generic 
  ancestor selector (table/list/content wrapper) if `containerSelector` doesn't 
  match, so it degrades gracefully across table and card/grid layouts.
- Settings (`priceSelector`, `containerSelector`, `defaultView`) are exposed via 
  `wp_localize_script` so site owners can adjust selectors without editing JS.

Validated on a staging site against both:
- Standard `[pmpro_levels]` shortcode output
- Advanced Levels Page Add-on output

### DIFF
--- a/blocks-shortcodes/pmpro-toggle-monthly-annual-pricing.php
+++ b/blocks-shortcodes/pmpro-toggle-monthly-annual-pricing.php
@@ -1,6 +1,10 @@
 <?php
 /**
  * Add a monthly/annual pricing toggle to the PMPro Levels Page and Advanced Levels Page shortcodes/blocks.
+ * 
+ * Style the toggle with CSS using the following classes: .pmpro-pricing-toggle (wrapper),
+ * .pmpro-pricing-toggle-btn (buttons), .pmpro-pricing-toggle-btn.is-active (active button),
+ * .pmpro-pricing-toggle-btn:not(.is-active) (inactive button).
  *
  * title: Add a Monthly/Annual Pricing Toggle to the PMPro Levels Page
  * layout: snippet-example
--- a/blocks-shortcodes/pmpro-toggle-monthly-annual-pricing.php
+++ b/blocks-shortcodes/pmpro-toggle-monthly-annual-pricing.php
@@ -1,0 +1,246 @@
+<?php
+/**
+ * Add a monthly/annual pricing toggle to the PMPro Levels Page and Advanced Levels Page shortcodes/blocks.
+ *
+ * title: Add a Monthly/Annual Pricing Toggle to the PMPro Levels Page
+ * layout: snippet-example
+ * collection: block-shortcode
+ * category: display
+ * link: TBD
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+add_action( 'wp_enqueue_scripts', 'my_pmpro_pricing_toggle_inline_assets' );
+function my_pmpro_pricing_toggle_inline_assets() {
+	global $post;
+
+	if ( is_admin() || ! is_a( $post, 'WP_Post' ) ) {
+		return;
+	}
+
+	// Detect either the default levels shortcode or the Advanced Levels
+	// Page Add-on's shortcode/block, anywhere on the site — not just a
+	// specific page slug, since PMPro's own docs note these can be placed
+	// on any page.
+	$has_levels_shortcode = has_shortcode( $post->post_content, 'pmpro_levels' )
+		|| has_shortcode( $post->post_content, 'pmpro_advanced_levels' )
+		|| has_block( 'pmpro-advanced-levels/advanced-levels-page', $post );
+
+	if ( ! $has_levels_shortcode ) {
+		return;
+	}
+
+	// Register a dummy handle so we have something to attach inline JS/CSS to.
+	wp_register_script( 'pmpro-pricing-toggle', '', array(), '1.0.0', true );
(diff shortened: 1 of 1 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/pmpro-snippets-library#501 — Add Monthly/Annual Pricing Toggle for PMPro Levels Page and Advanced Levels Page**
*MaryOJob → dev | 1 file, +250 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/501>

---

**Summary**
The WP integration plumbing is solid (correct dummy-handle inline script/style pattern, verified selectors against core — `.pmpro_level-price` and `.pmpro_levels_table` exist in `pages/levels.php:73,104`), but the core approach — regex-scraping the rendered, translated price string instead of using structured level data already available server-side — is fragile enough to break the feature outright for large classes of real sites, and one block leftover ships debug styling to production. Not ready to merge.

---

**Issues**

- **Critical** `pmpro-toggle-monthly-annual-pricing.php:84` — `renderNode()` wipes the entire price node (`node.textContent = ''`) and rebuilds it from only the matched annual figure. `pmpro_getLevelCost()` (`paid-memberships-pro/includes/functions.php:339`) can produce text like "$10 now and then $120 per Year" (differing initial payment), plus trial-period and finite-billing-limit clauses. All of that is discarded — the toggle silently drops the up-front payment, trial terms, and payment-count disclosures from the displayed price. Derive the monthly figure from the level's raw amounts (available server-side) and append it alongside the existing text rather than replacing the node's content.

- **Major** `pmpro-toggle-monthly-annual-pricing.php:60,68` — `ANNUAL_PATTERN` hardcodes the literal English word "year" and requires a leading `$`/`£`/`€`. The source text comes from `pmpro_translate_billing_period()` (`__("Year", 'paid-memberships-pro')`, `includes/localization.php:39-40`), so any non-English site never matches — `parsePrice()` returns null for every level and the toggle never appears, with nothing logged to explain why. Separately, checked `includes/currencies.php`: currencies like ALL, CZK, PLN, RON, RUB, SGD, and UAH use `position => 'right'` (symbol after the amount, e.g. `40,00 Kč`), and CZK/PLN/RON/RUB/UAH also use a comma decimal separator — both silently defeat this regex (SGD even uses a plain `$` but right-positioned, so the symbol-first assumption alone breaks it). Source the raw numeric amount and currency config via `wp_localize_script` instead of parsing formatted, translated display text — it sidesteps all of this at once.

- **Major** `pmpro-toggle-monthly-annual-pricing.php:31-33` — Detection covers the `pmpro_levels`/`pmpro_advanced_levels` shortcodes and the Advanced Levels Page block, but misses PMPro core's own Gutenberg block, `pmpro/levels-page` (confirmed in `blocks/src/levels-page/block.json`), whose `render.php` outputs the identical `pages/levels.php` template — same `.pmpro_level-price`/`.pmpro_levels_table` markup — but leaves no `[pmpro_levels]` shortcode text in `post_content` for `has_shortcode()` to catch. Add `has_block( 'pmpro/levels-page', $post )`. Note: this block isn't in the latest tagged core release (`v2.5.9.1`) yet, so it's not affecting live sites today, but it's already in core's source tree and the PR's own description claims coverage "regardless of site configuration."

- **Minor** `pmpro-toggle-monthly-annual-pricing.php:112-119,240-243` — Debug scaffolding shipped as-is: the JS comment literally says "Remove this classList.add line (and the matching CSS rule) once you're done testing," and the matching `.pmpro-pricing-toggle-highlight-row` outline rule is marked the same way. This is a copy-paste snippets library — nothing strips this before publication, so every site that installs the recipe gets a permanent dashed blue outline on its pricing table. Remove both, or gate behind an explicit debug flag that defaults off.

- **Minor** `pmpro-toggle-monthly-annual-pricing.php` — Levels pages mixing monthly- and annual-billed levels get an inconsistent result: `parsePrice()` only matches "year" text, so monthly levels never populate `dataset.pmproAnnualAmount` and are skipped by `renderAll()`. The toggle still renders (because at least one annual level matched) but switching to "Monthly" leaves the monthly levels' original text untouched next to the converted ones. Worth handling explicitly (store the monthly levels' own amount so both views stay consistent) or hiding the toggle when the page isn't uniformly annual.

---

**Looks Good**
The `wp_register_script('', ...)` + `wp_add_inline_script` pattern is the right idiom for attaching inline JS without a phantom asset file, and gating on `has_shortcode()`/`has_block()` rather than a page slug matches PMPro's own guidance that these can live on any page. Accessibility got real, unprompted attention: `aria-pressed`, an `aria-live` region, `role="group"`, and screen-reader text on the toggle.

---

**Questions**
1. Given the i18n/currency issues above, is there a reason to parse the rendered display string at all instead of localizing the raw `initial_payment`/`billing_amount` per level from PHP in the same `wp_enqueue_scripts` hook? That would remove the two Major findings entirely.
2. For the Advanced Levels Page add-on's `compare_table` layout, does the surrounding markup (`<td>` vs `<div>`) change how the fallback ancestor selector (`closest('table, ul, ol, ...')`) behaves? Worth a quick check against that template variant.

PR #509, opened 2026-08-28
Add snippet to display restricted PDFs inline in the browser

dwanjuki:
Adds `restricting-content/display-restricted-pdfs-inline-in-browser.php`.

PMPro serves restricted images inline and all other file types as downloads. This snippet filters `pmpro_restricted_file_content_disposition` (added in PMPro 3.6) to also serve PDFs inline, so members can read them in the browser instead of being forced to download.

The MIME type is read with `finfo` so only real PDFs are affected; anything else falls through to the disposition core already chose.

`link:` is set to `TBD` — there's no companion article for this recipe yet.

### DIFF
--- a/restricting-content/display-restricted-pdfs-inline-in-browser.php
+++ b/restricting-content/display-restricted-pdfs-inline-in-browser.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Display restricted PDFs inline in the browser instead of forcing a download.
+ *
+ * By default, Paid Memberships Pro serves restricted images inline and all other
+ * file types as downloads. This recipe also serves PDFs inline.
+ *
+ * Restrict files: https://www.paidmembershipspro.com/locking-down-protecting-files-with-pmpro/
+ *
+ * title: Display Restricted PDFs in the Browser
+ * layout: snippet
+ * collection: restricting-content
+ * category: content, restriction
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#509 — Add snippet to display restricted PDFs inline in the browser**
*dwanjuki → dev | 1 file, +32 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/509>

---

**Summary**
Ready to merge. Small, correct snippet — the filter signature and usage match core exactly (`repos/strangerstudios/paid-memberships-pro/includes/restricted-files.php:103`), and the non-PDF fallthrough preserves default behavior.

---

**Issues**
- **Minor** `restricting-content/display-restricted-pdfs-inline-in-browser.php:20-22` — Opens a second `finfo` resource to re-detect the MIME type. Core already does this one line earlier (`restricted-files.php:87-88`) but only passes the `wp_getimagesize()` boolean into the filter, not the MIME type itself — so there's no way for the snippet to avoid the extra call. This is a core API gap, not something to fix in this PR. No action needed.

---

**Looks Good**
- Non-PDF and undetectable files (`finfo_file()` returning `false`) correctly fall through to the original `$content_disposition` rather than forcing a value.
- MIME detection is content-based (`finfo`), not extension-based, so a `.pdf`-named non-PDF file won't get served inline.
- Core's post-filter validation (`restricted-files.php:104-105`) clamps any invalid return to `attachment`, so this snippet can't produce a broken `Content-Disposition` header even in a failure case.
- `link: TBD` matches the established pattern for recipes without a companion article — confirmed against other files in `restricting-content/` (e.g. `allow-access-for-admins.php:9`).

---

**Questions**
1. Worth a one-line note in the docblock that this trades download-only for browser-rendered PDFs? Modern browsers sandbox PDF viewers so it's low-risk, but some site owners may not expect the behavior change. Not blocking.

PR #515, opened 2026-09-11
Add snippet to customize cancellation email for failed payments

dwanjuki:
Adds a `{{ cancelled_due_to_payment_failure }}` email template variable to the Cancel and Cancel Admin emails, so the body can use Liquid syntax to show different content when a membership was cancelled because a recurring payment failed:

```
{% if cancelled_due_to_payment_failure %}
Content for members cancelled due to payment failure.
{% else %}
Content for members who initiated cancellation.
{% endif %}
```

### How it detects the cause

Core does not pass a cancellation reason to the cancel email, but every gateway routes recurring payment failures through `pmpro_handle_recurring_payment_failure_at_gateway()`, which saves a pending order and adds `last_failure_email_sent` order meta to it. A later successful retry reuses that same order and flips it to success, so an unresolved failure is simply an incomplete order carrying that meta.

Two details this depends on:

- `PMPro_Subscription::save()` marks incomplete orders as error when the subscription is cancelled, and that runs before the cancel email is sent, so both `pending` and `error` count.
- `get_subscriptions_for_user()` defaults to active subscriptions only, and the subscription is already cancelled by email time, so the status list is passed explicitly.

Cancellations requested by a member or an admin are excluded with an `is_user_logged_in()` check, since gateway webhooks are unauthenticated. Without it, a member who cancels while payments are still retrying would be told their membership ended because of a payment failure.

### Still to do

- `link:` in the header is `TBD` pending the article URL.

### Testing

Verified on Stripe: a subscription cancelled after retries were exhausted takes the payment-failure branch, and a member-initiated cancellation takes the else branch.

### DIFF
--- a/email/customize-cancellation-email-for-failed-payments.php
+++ b/email/customize-cancellation-email-for-failed-payments.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * Add a {{ cancelled_due_to_payment_failure }} email template variable to the
+ * PMPro Cancel emails so you can vary the wording with Liquid syntax when a
+ * membership was cancelled because a recurring payment failed.
+ *
+ * Usage in Memberships > Settings > Email Templates > Cancel:
+ *
+ *   {% if cancelled_due_to_payment_failure %}
+ *   Content for members cancelled due to payment failure.
+ *   {% else %}
+ *   Content for members who initiated cancellation.
+ *   {% endif %}
+ *
+ * title: Customize the Cancellation Email for Memberships Cancelled Due to Failed Payments
+ * layout: snippet
+ * collection: email
+ * category: emails, email templates
+ * link: TBD
+ *
+ * You can add this recipe to your site by creating a custom plugin
+ * or using the Code Snippets plugin available for free in the WordPress repository.
+ * Read this companion article for step-by-step directions on either method.
+ * https://www.paidmembershipspro.com/create-a-plugin-for-pmpro-customizations/
+ */
+
+/**
+ * Was this membership cancelled by the gateway after a payment failed?
+ *
+ * @param int $user_id  The ID of the member.
+ * @param int $level_id The cancelled membership level ID.
+ * @return bool Whether the cancellation was caused by a failed payment.
+ */
+function my_pmpro_cancelled_due_to_payment_failure( $user_id, $level_id ) {
+	if ( ! class_exists( 'PMPro_Subscription' ) ) {
+		return false;
+	}
+
+	// Bail if the cancellation was admin/member initiated.
+	if ( is_user_logged_in() ) {
+		return false;
+	}
+
+	// Get the most recent subscription for the cancelled level.
+	$subscriptions = PMPro_Subscription::get_subscriptions_for_user(
+		$user_id,
+		$level_id,
+		array( 'active', 'cancelled' )
+	);
+	if ( empty( $subscriptions ) ) {
+		return false;
+	}
+	$subscription = current( $subscriptions );
+
+	// A failed payment leaves a pending order. Cancelling then flips it to error
+	// in PMPro_Subscription::save(), before the cancel email, so accept both.
+	$failed_orders = $subscription->get_orders( array( 'status' => array( 'pending', 'error' ) ) );
+
+	foreach ( $failed_orders as $order ) {
+		// Only the failure handler writes this meta.
+		if ( get_pmpro_membership_order_meta( $order->id, 'last_failure_email_sent', true ) ) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+/**
+ * Add a {{ cancelled_due_to_payment_failure }} variable to the cancel emails.
+ *
+ * @param array      $data  The email template variables.
+ * @param PMProEmail $email The email object being sent.
+ * @return array The filtered email template variables.
+ */
+function my_pmpro_add_payment_failure_email_variable( $data, $email ) {
+	// Only affect the cancel emails. Remove 'cancel_admin' if you only want to change the email sent to the member.
+	if ( empty( $email->template ) || ! in_array( $email->template, array( 'cancel', 'cancel_admin' ), true ) ) {
+		return $data;
+	}
+
+	// Default to empty so the variable always exists and never falls through to usermeta lookup.
+	$data['cancelled_due_to_payment_failure'] = '';
+
+	// Bail if there is no membership level to check.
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/pmpro-snippets-library#515 — Add snippet to customize cancellation email for failed payments**
*dwanjuki → dev | 1 file, +105 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/515>

---

**Summary**
Ready to merge once the docblock states its gateway scope. The detection chain holds up against core: `pmpro_handle_recurring_payment_failure_at_gateway()` writes `last_failure_email_sent` (`includes/gateway-request-handlers.php:303`), `PMPro_Subscription::save()` flips pending orders to `error` before `sendCancelEmail()` runs in `pmpro_handle_subscription_cancellation_at_gateway()`, and a successful retry reuses the order and drops it out of the `pending`/`error` filter. The one real gap is that not every gateway goes through that failure handler, so the snippet should say which gateways it covers rather than present itself as universal.

---

**Issues**

- **Major** `email/customize-cancellation-email-for-failed-payments.php:59-64` — `last_failure_email_sent` is written only by `pmpro_handle_recurring_payment_failure_at_gateway()`. Stripe (`services/class-pmpro-stripe-webhook-handler.php:382`) and PayPal (`services/ipnhandler.php:681`) call it; Braintree does not. Braintree's webhook fires `pmpro_subscription_payment_failed` and `sendBillingFailureEmail()` directly (`services/braintree-webhook.php:216,259,294,337`) and later cancels through `pmpro_handle_subscription_cancellation_at_gateway()` (`:399`), so a Braintree failed-payment cancellation always takes the `{% else %}` branch and tells the member they cancelled. Fix in the docblock is enough for a snippet: state that it works for gateways that route failures through core's handler (Stripe, PayPal) and not Braintree. If you want full coverage, hook `pmpro_subscription_payment_failed` to set your own subscription meta and clear it on `pmpro_subscription_payment_completed`.

- **Minor** `email/customize-cancellation-email-for-failed-payments.php:35` — The guard checks `PMPro_Subscription` but the Liquid `{% if %}` syntax the docblock depends on requires the renderer added in PMPro 3.7 (`classes/class-pmpro-liquid-renderer.php` is `@since 3.7`). On 3.0–3.6 the variable is set but the template body renders the tags literally. Either add `class_exists( 'PMPro_Liquid_Renderer' )` to the guard or note the 3.7+ requirement in the header.

- **Minor** `email/customize-cancellation-email-for-failed-payments.php:59-68` — `get_orders()` has no recency bound, so any historical `error` order with the meta (a failure that was resolved out of band rather than via gateway retry on the same order) will flag a later, unrelated gateway-side cancellation as payment failure. The `is_user_logged_in()` gate removes the common false positive, but a one-line comment on the assumption, or checking only the most recent order, would make the limitation visible.

- **Minor** `email/customize-cancellation-email-for-failed-payments.php:91` — When multiple levels cancel at once, core passes only the first level ID as `membership_id` (`classes/email-templates/class-pmpro-email-template-cancel.php:159`). If the failed subscription is on another level the check misses. Rare, but worth a comment.

---

**Looks Good**
Defaulting `$data['cancelled_due_to_payment_failure']` to `''` before every early return is load-bearing: it stops the Liquid renderer's usermeta fallback from resolving the variable from the recipient's profile. Resolving the user from `user_login` rather than the recipient address is the right call, since `cancel_admin` goes to the admin. Strict `in_array()` on the template check, all query inputs int-cast or hardcoded, nothing read from request globals.

---

**Questions**
1. A member cancelling via the Stripe billing portal arrives as an unauthenticated webhook. With a pending failed order they get the payment-failure copy even though they initiated it. Acceptable as a documented limitation?

dwanjuki:
@flintfromthebasement Cancelling several levels at once sends one email per level rather than one email covering all of them

flintfromthebasement (approved):
**Follow-up review**

**PR: strangerstudios/pmpro-snippets-library#515 — Add snippet to customize cancellation email for failed payments**
*dwanjuki → dev | 1 file, +111 -0*
<https://github.com/strangerstudios/pmpro-snippets-library/pull/515>

**Summary**
All blocking concerns from my last pass are cleared. This is ready to merge.

**Prior issues — status**

- **Major** `last_failure_email_sent` only written by the core failure handler, so Braintree cancellations always hit the `{% else %}` branch — **Resolved.** New docblock section ("Gateway support: needs a gateway that uses PMPro's built-in handling...") states the Stripe/PayPal scope and documents the Braintree gap instead of presenting the snippet as universal. That's the right fix for a snippet.

- **Minor** Liquid `{% if %}` syntax requires PMPro 3.7+ — **Resolved.** Docblock now states "Requires Paid Memberships Pro v3.7+ for Liquid syntax in email templates."

- **Minor** `get_orders()` has no recency bound, so an unrelated resolved-out-of-band failure could mis-flag a later cancellation — **Resolved.** Added comment makes the assumption explicit ("A retry reuses the order, so one still incomplete is assumed unresolved.") — that's exactly the visibility I asked for; the assumption is reasonable for a snippet and now documented rather than silent.

- **Minor** Core passes only the first level ID as `membership_id` when multiple levels cancel at once, so a failed subscription on a different level would be missed — **Withdrawing.** Per @dwanjuki's comment, cancelling multiple levels sends one email per level, not one email covering all of them. I verified against `preheaders/cancel.php:73-114` upstream/dev: the member-cancel flow loops `foreach ($old_level_ids as $old_level_id)` and calls `sendCancelEmail($current_user, $old_level_id)` once per level. `sendCancelEmail()` itself (`classes/class.pmproemail.php:433-438`) explicitly `_doing_it_wrong()`s if passed an array — every current caller (`preheaders/cancel.php`, `gateway-request-handlers.php`, `twocheckout-ins.php`, member-edit admin panel) passes a single int. The array/"first level id" branch in `PMPro_Email_Template_Cancel::get_email_template_variables()` exists in code but isn't reachable through any current cancel-email call site, so `membership_id` will always correctly match the level this specific email is about. My original concern was based on a dead path — fair catch.

**New findings**
None. The diff since my last review only touches docblock comments (gateway scope note, 3.7+ requirement, recency-assumption comment) — no logic changed, nothing new to flag.

**Looks Good**
Good instinct verifying the multi-level claim against actual behavior rather than just taking the code at face value — the array branch really is unreachable in practice, and the docblock additions land exactly the caveats a snippet needs without overengineering the logic to handle Braintree.

